PR #71, opened 2026-05-08
[codex] Fix ZIP-317 shielding fee mismatch

Thunnini:
## Summary

Fixes a ZIP-317 fee mismatch in transparent-input transaction construction by using a wallet-local conservative ZIP-317 fee rule. The rule preserves the standard ZIP-317 parameters but rounds known transparent input sizes up to the standard P2PKH input size before delegating to `StandardFeeRule::Zip317`.

## Root Cause

For transparent-heavy shielding, proposal construction could price P2PKH inputs at the ZIP-317 standard 150-byte size while transaction building later used the exact serialized input size, such as 149 bytes. With 322 inputs, that lowered the builder fee by two marginal actions, leaving exactly 10,000 zats unaccounted for. Shielding transactions do not add change, so the builder failed with a `ZatBalance(10000)` change-required error.

## Changes

- Added `ConservativeZip317FeeRule` for wallet send/shield proposal construction.
- Updated stored proposal typing to use the wallet-local fee rule.
- Added a regression test that creates 322 transparent UTXOs, proposes shielding, and verifies transaction construction succeeds.

## Validation

- `cd rust && cargo fmt --check`
- `cd rust && cargo test many_utxo_shielding_builds_with_conservative_zip317_fee -- --nocapture`

The regression test now builds one transaction successfully with `proposal_fee=1630000` and no `ZatBalance(10000)` change-required error.

### DIFF
--- a/rust/src/wallet/sync/mod.rs
+++ b/rust/src/wallet/sync/mod.rs
@@ -5,7 +5,6 @@ use zcash_client_backend::{
         wallet::ConfirmationsPolicy,
         WalletCommitmentTrees, WalletRead, WalletWrite,
     },
-    fees::StandardFeeRule,
     proto::service::TreeState,
 };
 use zcash_client_sqlite::{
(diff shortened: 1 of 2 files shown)


Thunnini:
@codex review

PR #74, opened 2026-05-08
Accounts screen

piatoss3612:
## Summary

- Add the Accounts screen flow for viewing the active account, switching accounts, opening per-account actions, renaming accounts, changing profile pictures, and removing accounts.
- Add account removal UI states so destructive removals show when sync is being stopped before the wallet DB is mutated.
- Track the wallet's seed-anchor account explicitly and use that metadata to decide whether an account can be removed.
- Fix the reset-then-import sync failure by invalidating the cached wallet DB path after wallet reset.

## Account removal model

The wallet DB must always retain at least one seed-backed `Derived` account while any other accounts remain. The first software account is the seed anchor because `init_wallet_db(Some(seed))` pins the DB to that seed, and future seed-requiring librustzcash migrations need a relevant derived account to validate against.

Additional software accounts are imported through the UFVK/import path so multiple seeds can coexist in one DB. Those accounts carry derivation metadata, but librustzcash still treats them as `Imported`; it cannot use their seed for seed-requiring DB initialization or migration relevance checks.

Because of that constraint, deleting the last seed-anchor account while imported accounts remain would leave the DB in an imported-only state. That state is risky because a future seed-requiring migration may be unable to open or migrate the wallet. The UI therefore blocks removing the last seed anchor as long as other accounts exist.

When only one account remains, removing it is treated as a full wallet reset instead of an account-list deletion. At that point there are no remaining accounts to preserve, so the safe behavior is to delete the wallet DB/keychain state and return the app to onboarding.

## Reset / sync DB path fix

Wallet DB filenames are stored in secure storage as `zcash_wallet_db_name` and include a random suffix. A wallet reset deletes secure storage, so the next create/import flow generates a new DB filename.

The sync provider cached the previous resolved DB path. After a reset followed by a new import, import/migration correctly initialized the new DB, but sync could still open the old cached path. SQLite creates a blank file when opening a missing path, so sync then failed with missing schema errors like `no such table: blocks`.

This PR clears the sync provider's cached DB path after wallet reset and routes destructive reset through the same sync-pause guard used by account removal. The guard also pauses stale sync work even when the account list is already empty, which covers reset/import transition windows.

## Validation

- `fvm flutter test test/app_bootstrap_test.dart test/providers/account_models_test.dart test/features/accounts/accounts_screen_test.dart`
- `cd rust && cargo test test_delete_account_removes_account_from_wallet_db`
- `fvm flutter test test/providers/wallet_mutation_guard_test.dart test/features/accounts/accounts_screen_test.dart`
- `fvm flutter analyze lib/src/providers/sync_provider.dart lib/src/providers/wallet_mutation_guard.dart lib/src/features/accounts/screens/accounts_screen.dart lib/src/features/onboarding/lost_password_screen.dart lib/src/providers/account_provider.dart test/providers/wallet_mutation_guard_test.dart test/features/accounts/accounts_screen_test.dart`


piatoss3612:
@codex review

PR #77, opened 2026-05-11
[codex] Restore Keystone QR support

Thunnini:
## Summary

Restores Keystone hardware wallet support with a QR-only flow.

- Adds Keystone onboarding under `/onboarding/keystone`, including account QR scan, account selection, wallet birthday input, and first-wallet password setup.
- Adds Keystone send support through the PCZT animated QR signing flow, signed QR scan, broadcast status handling, and Sapling params download prompting when needed.
- Adds Keystone transparent shielding support using the same QR confirmation and signature scan flow.
- Removes the Rust USB transport path and updates the scanner/camera UI around `mobile_scanner`.
- Handles UR scan session reset/noise cases, post-broadcast local-storage warning states, and hardware-account restrictions such as unavailable secret passphrase viewing.

## Why

Keystone support had been removed, but the product now needs Keystone accounts, send signing, and transparent shielding restored without USB. The new flow is camera + QR only and follows the current onboarding/send screen design direction.

## Impact

Users can import Keystone accounts as first or additional wallet accounts, scan animated Keystone account/signature QRs, send from Keystone accounts, and shield transparent balance for Keystone accounts. Software account flows remain on the existing mnemonic signing path.

## Validation

- `fvm dart analyze lib`
- `cd rust && cargo check`
- `git diff --check 67798c586f54b2ca5663914092d592de10b31c02`


Thunnini:
@codex review

PR #92, opened 2026-05-13
Keep seed derivation inside Rust

### LINKED ISSUE #86: [Audit] Keep seed derivation and signing out of Dart heap
## Finding

Software send and shielding export raw seed bytes into Dart-managed memory, then pass them back to Rust.

Evidence:
- `rust/src/api/wallet.rs:259` exposes `derive_seed` as `Vec<u8>`.
- `lib/src/rust/api/wallet.dart:133` exposes it as `Future<Uint8List>`.
- `lib/src/features/send/screens/send_status_screen.dart:390` derives seed bytes for send.
- `lib/src/features/home/screens/home_screen.dart:125` derives seed bytes for shielding.

## Impact

The wallet root seed lives in Dart heap memory where lifetime and zeroization are much harder to control. That increases exposure to crash dumps, debugging tools, accidental retention, and future UI-layer misuse.

## Proposed fix

Move mnemonic-to-seed derivation and signing entirely behind Rust APIs. Prefer APIs that take the mnemonic only at the Rust boundary and immediately wrap/zeroize temporary secret material, or a broader Rust-side signing entry point that never returns raw seed bytes.

## Tests

Add an API-level regression that software send and shielding still work after removing UI calls to `deriveSeed`, plus a static grep-style test or review guard that production Dart no longer calls `deriveSeed` for signing.

zmanian:
## Summary

- remove the public `derive_seed` / `deriveSeed` FRB API so Dart no longer requests raw 64-byte seed material
- pass mnemonics into the send and transparent-shielding Rust API entry points, deriving seed bytes inside Rust immediately before signing
- regenerate FRB bindings and add a guard test that fails if production feature code starts calling `deriveSeed(` again

Closes #86.

## Verification

- `fvm flutter test test/security/no_dart_seed_derivation_test.dart`
- `fvm flutter analyze lib/src/features/home/screens/home_screen.dart lib/src/features/send/screens/send_status_screen.dart test/security/no_dart_seed_derivation_test.dart`
- `cargo test` from `rust/`

### DIFF
--- a/lib/src/features/home/screens/home_screen.dart
+++ b/lib/src/features/home/screens/home_screen.dart
@@ -24,7 +24,6 @@ import '../../../providers/rpc_endpoint_failover_provider.dart';
 import '../../../providers/sync_provider.dart';
 import '../../../providers/wallet_provider.dart';
 import '../../../rust/api/sync.dart' as rust_sync;
-import '../../../rust/api/wallet.dart' as rust_wallet;
 import '../../activity/activity_row_mapper.dart';
 import '../../activity/models/activity_row_data.dart';
 import '../../activity/screens/activity_transaction_status_screen.dart';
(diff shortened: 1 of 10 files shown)


Thunnini:
Closing as superseded by #105. That PR covers the same goal of keeping seed derivation inside Rust and extends it by avoiding Dart String mnemonic transport in the runtime send/shield paths and zeroizing mnemonic/seed buffers earlier.

PR #115, opened 2026-05-14
[codex] Clear create onboarding mnemonic after setup

Thunnini:
## Summary

Clears the create-onboarding mnemonic provider and revealed flag after first-wallet password setup completes successfully.

## Why

The create onboarding flow preserved the generated mnemonic while navigating between onboarding screens, but the set-password success path created the wallet and routed to `/home` without clearing that root-scoped state. This could keep a plaintext mnemonic and revealed state alive longer than intended.

## Changes

- Added a shared `clearCreateOnboardingSecretState` helper for the create-onboarding mnemonic and reveal flag.
- Reused the helper in the existing intro and direct-create cleanup paths.
- Called the helper after successful `SetPasswordFlow.create` setup before navigating home.
- Added focused coverage for the shared cleanup helper while preserving pending mnemonic reuse during in-progress onboarding.

## Validation

- `fvm flutter test test/features/onboarding/create_onboarding_mnemonic_test.dart`
- `fvm flutter analyze`

### DIFF
--- a/lib/src/features/onboarding/create/intro_zcash_screen.dart
+++ b/lib/src/features/onboarding/create/intro_zcash_screen.dart
@@ -26,10 +26,7 @@ class _IntroZcashScreenState extends ConsumerState<IntroZcashScreen> {
     super.initState();
     WidgetsBinding.instance.addPostFrameCallback((_) {
       if (!mounted) return;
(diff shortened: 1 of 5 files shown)


Thunnini:
@codex review

PR #132, opened 2026-05-21
feat(zns): Zcash Name System integration

craftsoldier:
Hey Thunini! Here's what we built 👋

## What's included

- **Forward lookup in the send screen** - type `alice.zcash`, get a Zcash Unified Address
- **Reverse lookup in the activity detail screen** - sent transactions show the recipient's ZNS name above their address
- **ZNS identity easter egg** - rename your account to your ZNS name, your profile picture background turns gold

---

## What we changed

- `send_screen.dart` - ZNS resolution wired into the address field, `_ZnsStatus` enum, debounce, staleness guard, tappable resolved name label
- `send_review_screen.dart`, `send_status_screen.dart` - `resolvedName` threaded through `SendReviewArgs` via `recipientLabel`
- `activity_transaction_status_screen.dart` - reverse lookup on sent transactions, name shown above address in "To" block
- `zns_provider.dart` - `ZnsResolver` with `resolveName`, `reverseResolve`, `isOwnedByAccount`, session-scoped `_nameCache`
- `rust/src/wallet/keys.rs` - `is_address_from_account`, O(1) Orchard FVK ownership check
- `rust/src/api/wallet.rs` - FRB wrapper exposing it to Dart
- `accounts_screen.dart`, `settings_screen.dart` - `_checkZnsEasterEgg` fires after successful rename
- `profile_pictures.dart`, `app_profile_picture.dart` - `zns-` prefix support, gold background `Color(0xFFFFB800)`

---

## What we did not change

- `AccountNameModal` - untouched, no new parameters
- Received transactions - sender is private in shielded Zcash, nothing to look up
- Send flow - proposal, fee estimation, and broadcast unchanged

---

## How we did this

**Forward lookup** triggers on a conjunction: Rust `validateAddress` rejects the input AND the input matches a ZNS protocol identifier (`.zcash`, `.zec`). Rust fires first as the authoritative local validator - ZNS only activates on explicit rejection. Full input is lowercased before the check so `Alice.ZCASH` and `alice.zcash` are equivalent. The existing `_addressSeq` staleness guard covers ZNS too.

**Reverse lookup** uses `ZnsResolver.reverseResolve` - calls `resolveAddress` on the SDK, takes the earliest-registered name, appends `.zcash`. Session-scoped `_nameCache` on the provider means repeated lookups for the same address are instant. `containsKey` rather than null check since null is a valid cached value (address has no registration).

**Ownership check** - `is_address_from_account` in Rust extracts the diversifier `d` from the UA's Orchard receiver and verifies `[ivk] × d == pk_d` against the account's Orchard FVK. O(1), no iteration over diversifier space, works for any diversified address.

**Easter egg** - `zns-` prefixed `profilePictureId` stored in secure storage. `AppProfilePicture` renders gold background when prefix is present, strips it before asset lookup so the avatar is unchanged.

### DIFF
--- a/lib/src/core/profile_pictures.dart
+++ b/lib/src/core/profile_pictures.dart
@@ -60,11 +60,19 @@ const kProfilePictureOptions = <ProfilePictureOption>[
   ),
 ];
 
+const _kZnsPrefix = 'zns-';
+
+bool isZnsProfilePictureId(String id) => id.startsWith(_kZnsPrefix);
+
+String _stripZnsPrefix(String id) =>
+    id.startsWith(_kZnsPrefix) ? id.substring(_kZnsPrefix.length) : id;
+
 ProfilePictureOption? findProfilePictureOption(String id) {
+  final baseId = _stripZnsPrefix(id);
   for (final option in kProfilePictureOptions) {
-    if (option.id == id) return option;
+    if (option.id == baseId) return option;
   }
-  if (_legacyKnightIdPattern.hasMatch(id)) {
+  if (_legacyKnightIdPattern.hasMatch(baseId)) {
     return kProfilePictureOptions.first;
   }
   return null;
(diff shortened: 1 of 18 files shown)


craftsoldier:
Marking this as a draft for now.

Before we move forward, we need to make sure this PR properly covers address book support and related account/contact management concerns. Converting to draft so the remaining work can be completed and reviewed without blocking the queue.

PR #172, opened 2026-06-02
Use shared last-moment voting policy

czarcas7ic:
Summary
- Pin zcash_voting to the shared last-moment timing helper commit.
- Route Vizor's last-moment buffer and single-share decisions through Rust wrappers.
- Recompute single-share mode immediately before commitment building and use the shared buffer during share planning.

Validation
- fvm flutter analyze
- cd rust && cargo check --locked --lib
- fvm flutter test test/providers/voting/voting_providers_test.dart test/features/voting/voting_status_screen_test.dart

Stacked on the current voting UI PR branch (`voting-ui`). Depends on the shared zcash_voting helper branch for last-moment timing.


czarcas7ic:
@codex review

PR #178, opened 2026-06-02
Prewarm voting power before poll detail

czarcas7ic:
Adds a display-only voting power provider that runs the light eligibility bundle setup without resolving PIR or marking delegation preparation complete.

The polls list now warms voting power for the first actionable active/voted round and gives tap navigation a bounded wait, so proposal detail usually opens with the round power already available while still routing if sync is slow.

Tests: fvm flutter analyze --no-pub; fvm flutter test --no-pub; git diff --check origin/voting-ui...HEAD.


czarcas7ic:
@codex review

PR #190, opened 2026-06-03
fix(voting): reset vote-tree cache before pre-sync failover

p0mvn:
## Summary

When the first pre-sync vote server fails after `syncVoteTree` has created or advanced the cached vote-tree client, the pre-sync failover path now calls `resetVotingSessionState` before retrying the next server. This matches the interactive `_syncVoteTreeWithFailover` behavior and avoids later VAN witness / vote-casting work reusing partial process-local state from the failed node.

## Test plan

- [x] `fvm flutter test test/providers/voting/voting_providers_test.dart --plain-name "vote tree pre-sync retries failover servers"`
- [x] `fvm flutter test test/providers/voting/voting_providers_test.dart`


czarcas7ic:
@p0mvn am closing as it was addressed, please re-open if I am wrong

PR #192, opened 2026-06-03
Fix guarded voting submission lifecycle

czarcas7ic:
Skip lock-time voting process cleanup while the active account has a guarded vote submission, and enforce the wallet sync max-wait budget so a stuck sync surfaces a retryable error instead of leaving a submission guard active indefinitely.

PIR warm-up treats the sync timeout as a non-fatal cache miss and clears the temporary waiting state.

Validation: fvm flutter test test/providers/account_provider_test.dart; fvm flutter test test/providers/voting/voting_providers_test.dart. fvm flutter analyze still reports the existing control_flow_in_finally lint in lib/src/features/voting/screens/voting_submission_confirmation_screen.dart.


czarcas7ic:
@codex review

PR #194, opened 2026-06-03
Gate voting with eligibility check

czarcas7ic:
Uses the voting crate eligibility helper to gate voting before bundle setup while preserving the reviewed ineligible-poll UI. Ineligible accounts can read poll options, see zero voting power, and get the same not eligible dialog from disabled option taps or submission controls.


czarcas7ic:
@codex review

czarcas7ic:
@codex review

PR #212, opened 2026-06-11
Redesign swap composer, review/status flow, and swap activity rows

piatoss3612:
## Summary

- Redesigns the swap composer to the new swap widget spec: serif (Young Serif Medium, lining figures) amount display, pinned page layout with bottom CTA, and restyled asset/slippage/address/QR-scan modals on the shared modal card.
- Redesigns the swap review, in-progress status, deposit, and failed surfaces to the new review frames, reusing the core review primitives from the send-review redesign (`ReviewInfoRow`, `ReviewListRow`, `ReviewWrapCard`).
- Absorbs the swap-leg ZEC transactions into the swap activity row as result sub rows (tappable for the received ZEC leg, inert for external legs) with a light entrance animation; the standalone duplicate rows are suppressed via byte-order-normalized txid matching.
- Unifies all swap detail screens (review / status / deposit / failed) on the shared `AppPaneScrollScaffold` + `AppPaneToolbar` pane model: pinned 48px back-link band, full-height 6px overlay scrollbar, no per-screen scroll chrome.
- Promotes reusable pieces into core: `AppModalCard`, `appSurfaceShadow`, `appSerifDisplayStyle`, and `ReviewInfoRow`/`ReviewListRow` extensions (`valueFit`, `copyText`, `trailingActionKey`).

## Behavior

- In-progress swap status pages are titled `Swap in progress...`; activity rows keep `Swapping...`.
- The completed swap's deposit tx row links to the NEAR Intents explorer; terminal timestamps use `completedAt ?? createdAt`.
- The NEAR Intents attribution now appears only on the composer; the redesigned detail pages drop it.
- The price-protection row stays removed (duplicate of slippage tolerance, per earlier copy audit); audited app copy intentionally overrides Figma copy throughout.
- The deposit-timeout/failed page ships the new knight illustration (single artwork for both themes).
- The status progress page is pinned to fit without a scrollbar at the 1080x720 reference window (regression-guarded); the review page scrolls through the shared scaffold when needed.

## Review Notes

- Based on `rowan/redesign-2-send-review` (#211) intentionally — the swap surfaces consume the core review primitives introduced there; a `main` base would pull that whole diff in.
- `269adfbf` widens `ReviewInfoRow`'s serif to the Young Serif Medium cut (w500 + `case` lining figures), which also restyles the send review/status amounts — this matches the Figma serif spec that send screens share.
- The Figma status frame nominally adds a 12px inset above the tabs, but the design frame overflows its own 720px window; the inset is intentionally dropped so the progress tab fits without scrolling (commented in code).
- Known pre-existing failure on the base branch (`test/widgetbook/send_review_status_use_cases_test.dart: failed use case keeps the dark card in the light theme`) reproduces on the clean base tip and is untouched here.

## Validation

- `fvm flutter analyze`
- `fvm flutter test` (full suite: only the pre-existing base-branch widgetbook failure above)
- Per-commit verification: each commit was built and tested in isolation during history construction
- Rendered review/status/deposit/failed at 1080x720 with real fonts and compared against the Figma frame renders


piatoss3612:
Verified the claim — two corrections, then addressed in f473232f:

1. **This wasn't introduced by this PR.** The pre-redesign review screen scoped the balance to the active account the same way (`691c70c0`, lines 129-130); `reviewAccountUuid` was only used for the account label/avatar there. The layout diff moved the surrounding code, which made the block look new.
2. **The mismatch scenario isn't reachable.** `swapStateProvider` listens to `accountProvider` and runs `_clearAccountScopedTransientState()` on every active-account change, which sets `reviewVisible: false` — and this screen redirects to `/swap` whenever `reviewVisible` is false. The listener fires before the widget rebuilds, so the screen never renders content while the active account differs from `reviewAccountUuid`. The start-in-flight edge is covered by `startIntent()`'s own account guard.

That said, pinning the gate to the quoted account is strictly more defensive, so f473232f scopes the balance to `reviewAccountUuid ?? activeAccountUuid` with a comment documenting the invariant.

piatoss3612:
Verified — the principle is correct, and the reachable case is actually broader than the example. Addressed in d5493b7f.

- The exact `expired` + recorded-deposit scenario is nearly unreachable: `expired` only arises from the local deadline resolution (`_resolveDepositDeadlineStatus`), which is suppressed whenever `swapHasConfirmedDepositEvidence` holds — i.e. any recorded `depositTxHash` that isn't `pendingBroadcast` blocks the expiry flip, and a `pendingBroadcast` deposit normally has no wallet transaction to absorb (broadcast-before-store). Only a broadcast-status race could line both up.
- However, **`refunded` hits the same problem through the normal path**: deposit lands on-chain, the provider refunds, and the swap row renders its amount unsigned (`includeSign: !(returnsFunds || timedOut)`) — so absorbing the Sent row left a refund credit in the feed with no matching debit.

d5493b7f gates the deposit-leg absorption on the row carrying the signed outgoing amount (`swapActivityRowAbsorbsDepositLeg`): refunded and expired intents keep their standalone Sent row, in-flight/completed swaps still absorb the duplicate. Covered with mapper unit tests plus two feed-level widget tests (refunded keeps `Sent`, in-flight suppresses it).

PR #223, opened 2026-06-13
Open manual entry when the import slot grid is tapped

dogemos:
## Summary
- The 24-slot numbered card reads as an input surface, so users tap it expecting to type — that tap now routes into the word-by-word manual wizard, same as the "Enter manually" link
- Active while the card is empty or showing a rejected paste; once a valid phrase fills the slots the card stays a plain review surface
- Adds a semantics button node ("Enter secret phrase manually") for screen readers

Fixes VZR-71

## Test plan
- New widget tests: tapping the empty grid opens the manual wizard; the grid stays tappable after a rejected paste (mobile lane, passing)


dogemos:
@codex review

PR #227, opened 2026-06-13
Keep the birthday date picker steady while browsing months

dogemos:
## Summary
- Month navigation bounced the pickers because their height tracked the 5- vs 6-week month layouts
- Native iOS sheet (`DatePickerHandler.swift`): the custom detent now resolves monotonically — grows once for a 6-week month (animated via `animateChanges`) and never shrinks back — instead of `invalidateDetents()` on every layout pass
- Flutter fallback calendar (Android + failure path): fixed six-week day grid, and the month/year grids render inside the same fixed-height box, so paging and mode switches never reflow the panel

Fixes VZR-75

## Test plan
- New `import_birthday_calendar_panel_test.dart`: panel height pinned across 5↔6-row month paging, day/month/year mode switches, and year paging (default lane, passing; the panel is shared with desktop)
- Swift side verified by building and running on the iOS simulator; manual check: birthday step → calendar → chevron July→August 2026 (5→6 rows) — sheet must not bounce

### DIFF
--- a/ios/Runner/DatePickerHandler.swift
+++ b/ios/Runner/DatePickerHandler.swift
@@ -151,6 +151,11 @@ private final class DatePickerSheetViewController: UIViewController,
   private let maxDate: Date
   private var pickedDate: Date?
   private var finished = false
+  /// Largest calendar fitting height seen so far. The detent is
+  /// monotonic — it grows for 6-week months and never shrinks back for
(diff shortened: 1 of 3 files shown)


dogemos:
Please double check Android since I only have a iOS device.

dogemos:
@codex review

PR #294, opened 2026-07-01
Fix mobile swap Keystone signing flow

piatoss3612:
## Summary

- Extract the mobile Keystone PCZT QR/signature scan round trip into a shared signing flow.
- Refactor mobile send signing to use that shared flow without changing its route contract.
- Add a mobile swap Keystone signing route for ZEC deposit swaps, matching the send-style full-screen QR -> scan flow.
- Prevent cancel/back navigation while a signed PCZT is being finalized so a broadcast cannot succeed without the activity record receiving the tx hash.

## Root Cause

Mobile swap was reusing the desktop-style Keystone signing overlay. That overlay opens `/send/keystone/scan` for signatures, but the mobile router does not expose that desktop scan route. Send already uses a separate full-screen mobile signing route, so swap needed the same mobile signing flow boundary instead of the desktop modal path.

## Commits

1. `Share mobile Keystone signing flow`
2. `Add mobile swap Keystone signing route`

## Validation

- `fvm flutter analyze`
- `fvm flutter test --tags mobile --run-skipped --dart-define=VIZOR_FORM_FACTOR=mobile test/features/swap/mobile_swap_keystone_signing_test.dart`
- `fvm flutter test --tags mobile --run-skipped --dart-define=VIZOR_FORM_FACTOR=mobile test/core/navigation/mobile_routes_test.dart`

## Notes

- Tested on a physical iPhone with a profile mobile build because iOS cannot launch Flutter debug builds from the home screen without Flutter tooling or Xcode.

### DIFF
--- a/lib/src/core/navigation/mobile_routes.dart
+++ b/lib/src/core/navigation/mobile_routes.dart
@@ -28,6 +28,7 @@ import '../../features/swap/screens/mobile/mobile_swap_screen.dart';
 import '../config/swap_feature_config.dart';
 import '../layout/mobile/app_mobile_shell.dart';
 import '../layout/mobile/app_mobile_tab_bar.dart';
+import '../theme/app_theme.dart';
 import '../widgets/app_icon.dart';
 import 'mobile_tab_history.dart';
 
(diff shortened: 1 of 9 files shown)


piatoss3612:
@codex review

PR #360, opened 2026-07-27
Persist transparent enhancement progress

ValarDragon:
## Summary

- persist completed transparent-address enhancement ranges in the wallet DB
- advance the checked-height cursor only after the lightwalletd stream
  completes successfully
- cover the exclusive range-end conversion and notification failures

## Root cause

`TransactionsInvolvingAddress` requests were serviced without calling
`WalletWrite::notify_address_checked` when the requested range completed.
The wallet DB therefore kept generating the same transparent history requests
on later syncs, causing established wallets to repeatedly download and process
historical transactions after restart or resync.

## Impact

Completed transparent-history ranges are now recorded, so subsequent syncs
continue incrementally instead of reprocessing the same history. Partial or
failed streams do not advance the cursor and remain safe to retry.

## Validation

- `cargo fmt -- --check`
- `cargo test sync_engine::enhance::tests` (12 passed)
- `cargo test` (438 passed, 1 ignored; Docker regtests ignored as designed)

No Rust or Dart API surface changes.

### DIFF
--- a/rust/src/wallet/sync_engine/enhance.rs
+++ b/rust/src/wallet/sync_engine/enhance.rs
@@ -23,7 +23,10 @@
 //! (e.g. a newly-decrypted transaction may reveal additional parent
 //! transactions to enhance).
 
-use std::collections::{BTreeMap, HashSet};
+use std::{
+    collections::{BTreeMap, HashSet},
+    fmt::Display,
+};
 
 use tonic::{transport::Channel, Code, Status};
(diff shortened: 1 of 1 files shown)


ValarDragon:
(this is vibecoded and unreviewed, but affects a user report of slow sync when they haven't logged in in 2 months, but very fast on direct usage of mnemonic)

PR #377, opened 2026-07-28
fix(send): reduce routine payment action fingerprint

ValarDragon:
## Summary

- use a single change output for ordinary payment proposals and fee estimates
- keep librustzcash's default two-action Orchard/Ironwood bundle padding
- retain the existing four-note management target for transparent shielding
- leave migration transaction construction unchanged

## Why

[CipherScan's wallet analysis](https://cipherscan.app/privacy/wallets) currently describes Vizor as producing "4 actions always" while grouping ZODL and other librustzcash wallets under a "2 actions (min)" baseline. The existing four-note change-management target can make a routine Vizor payment create extra change outputs while the account is below that target, producing a more distinctive action-count shape.

This change makes ordinary payments use `SplitPolicy::single_output()`. A common one-recipient/one-change transaction can therefore stay at the standard two-action floor. Transactions with multiple inputs, recipients, or cross-pool components may still legitimately require more than two actions.

Important source-level nuance: this is a Vizor fingerprint-reduction choice, not strict ZODL parity. The current official ZODL SDKs also configure a target of four sufficiently valued wallet notes:

- [ZODL Android SDK change policy](https://github.com/zcash/zcash-android-wallet-sdk/blob/6b5ba375c457257e28ec2f971dc89c6402ea241e/backend-lib/src/main/rust/lib.rs#L2094-L2097)
- [ZODL Swift SDK change policy](https://github.com/zcash/zcash-swift-wallet-sdk/blob/f51ed74aa70df959eee7af501f0f920438abe07d/rust/src/lib.rs#L2167-L2170)

CipherScan appears to compare Vizor's change-splitting behavior with the SDK family's lower-level two-action bundle minimum. This PR deliberately removes proactive splitting only from ordinary Vizor payments, which avoids relying on that ambiguous comparison.

## Testing

- `cargo test note_management` — 2 passed
- `cargo test --lib` — 472 passed, 1 ignored
- `cargo fmt --all -- --check`

### DIFF
--- a/rust/src/wallet/sync/send.rs
+++ b/rust/src/wallet/sync/send.rs
@@ -115,6 +115,8 @@ use super::{
 
 const UNBROADCAST_MIGRATION_RECOVERY_SAFETY_BLOCKS: u32 = 10;
 const SEND_PROPOSAL_LOCK_BLOCKS: u32 = 40;
+const NOTE_MANAGEMENT_TARGET_OUTPUT_COUNT: usize = 4;
+const MIN_NOTE_MANAGEMENT_OUTPUT_ZATOSHI: u64 = 10_000_000;
 
 fn send_proposal_lock_expiry(min_target_height: BlockHeight) -> BlockHeight {
     min_target_height + SEND_PROPOSAL_LOCK_BLOCKS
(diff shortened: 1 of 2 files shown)


czarcas7ic:
@ValarDragon closing given the tool was updated and my take is that there is no actual issue here (its possible when he made the tool, there was in fact this difference)

PR #380, opened 2026-07-28
DNM: Draft: Allow active migrations to adopt shorter timing

czarcas7ic:
Adds an opt-in “Use shorter timing” action on desktop and mobile for legacy active migrations, backed by an atomic Rust retiming API. The change redraws only unsubmitted work under the 90-minute policy while leaving submitted transactions untouched.

Before mutation, iOS background delivery is paused, accepted receipts are reconciled, and uncertain submissions fail closed. Existing signed transactions are reused only when their canonical ZIP 318 expiry remains valid; otherwise Vizor requests a fresh signature. This PR is stacked on the shorter migration timing branch.


czarcas7ic:
Draft generated code, not ready for review.

PR #381, opened 2026-07-29
Route migration transactions through a separate relay

czarcas7ic:
## Summary

Routes newly created Ironwood migration transactions through a separate Zebra-compatible `sendrawtransaction` relay when the wallet uses a recognized built-in lightwalletd preset. This covers both denomination-preparation transactions and scheduled Phase 2 Orchard-to-Ironwood transactions, preventing the lightwalletd server from learning their transaction IDs and linking them to the migration from the same client.

Managed mainnet configurations use an explicit build-time relay override when present; otherwise they use the HTTPS submission-only endpoint at `https://zakura-broadcast.valargroup.dev`. Testnet builds can opt into `https://zakura-broadcast.testnet.valargroup.dev` with `VIZOR_ZCASH_TRANSACTION_RELAY_URL_TEST`. Custom lightwalletd configurations retain the existing flow, including custom entries whose URL matches a built-in preset. Existing and in-progress migration runs remain on lightwalletd; only runs created after this change can opt into the separate relay.

## Design notes

The selected submission target is stored immutably on the migration run and reused for preparation and Phase 2. Relay transactions are validated and stored locally, excluded from generic lightwalletd resubmission, and enhanced from their local bytes plus compact-block scan state. Confirmation and expiry therefore do not require a transaction-specific lightwalletd request.

The iOS background outbox persists the same relay selection for scheduled Phase 2 submissions while continuing to use lightwalletd only for chain-tip data. Managed relays use HTTPS and expose transaction submission without general chain RPCs. Relay requests reject redirects, URL credentials, query strings, and fragments; stream responses through a 64 KiB cap; and verify the returned transaction ID.


czarcas7ic:
@codex review

rust/src/wallet/sync/send.rs
```diff
@@ -4516,7 +4587,29 @@ async fn broadcast_pending_denomination_stages(
...
-        if let Err(e) = broadcast_raw_transaction(&mut client, &stage.raw_tx).await {
+        let broadcast_result = if let Some(client) = relay_client.as_ref() {
+            match validate_separate_relay_denomination_stage(stage).and_then(|()| {
+                decrypt_and_store_migration_tx(db_path, network, &stage.raw_tx).map_err(|e| {
+                    format!("Store denomination split before separate-relay submission: {e}")
+                })
+            }) {
+                Ok(()) => {
+                    client
+                        .send_raw_transaction(&stage.raw_tx, &stage.expected_txid_hex)
```
p0mvn: For separate-relay denomination stages this PR does 

```
validate → decrypt_and_store → send_raw_transaction
```

while the lightwalletd path still does broadcast then store.


**Recommendation**: flip the relay path to match lightwalletd:

- validate_separate_relay_denomination_stage
- send_raw_transaction (keep duplicate-equivalent as success)
- decrypt_and_store_migration_tx (reuse existing store-failure retry messaging)
- mark_denomination_stage_broadcasted

If the relay permanently rejects after a successful store, notes can already look spent while the stage stays PENDING_BROADCAST. Expiry/NotInMainChain may eventually recover, but that is a worse intermediate failure than the LWD path and conflicts with the project’s PCZT “broadcast before persist” rule.

p0mvn (commented):
I think that users should also be able to disable/toggle this behavior without having to set custom endpoint override.

A clear setting (or advanced toggle) like “Submit migration preparation via separate relay,” default on.

rust/src/wallet/sync/send.rs
czarcas7ic: Order changed here https://github.com/chainapsis/vizor-wallet/pull/381/commits/e3533ea6650c10b3b7e0707290356e3628e56ead

p0mvn (approved):
I think we could also add an intermediary step similar to #388 down the road

czarcas7ic:
We are currently relying on a single URL for tx submission in the prep (split) transactions. We should update this to also use LWD endpoints (that will involve ensuring that if we use a LWD endpoint that it doesn't match the same provider as the one I am syncing with)

Another issue here is that we are still using LWD for the migration tx which is the big one, I am updating to fix this.

Lastly, once we do have more endpoints, we can probably decided to submit to 2 out of X instead of just 1, but that can be punted.

PR #383, opened 2026-07-29
Shorten migration timing and anchor age range

ValarDragon:
## Summary

- change preparation spacing from a 24-block mean (30 minutes) to a
  16-block mean (20 minutes)
- keep the preparation maximum at 96 blocks (2 hours)
- change migration transfer spacing from a 72-block mean to a 66-block mean
- keep the migration transfer maximum at 576 blocks (12 hours)
- reduce the ZIP-318 anchor age cap from 16 buckets to 4 buckets

## Anchor age integration

Main now includes `a101ef38` (`Allow latest ZIP-318 anchor bucket`), which
versions new migrations onto the age-zero anchor policy while preserving the
policy of existing runs. This branch builds on that implementation and narrows
the shared maximum age: the latest-anchor policy selects from ages 0 through 4,
while legacy age-one policies select from ages 1 through 4. The existing
per-age weighting is unchanged; ages 5 through 16 are removed.

Related ZIP update:
[zcash/zips#1343 — ZIP 318: Shorten migration timing](https://github.com/zcash/zips/pull/1343).

## Why 66 blocks

The sampler redraws exponential delays above the 576-block cap. A 66-block
parameter has an expected sampled delay of about 65.91 blocks, versus about
133.28 blocks for the legacy 144-block policy. This is an effective 2.02x
speedup while retaining the existing 12-hour maximum.

At the 75-second target block spacing, the expected transfer interval is about
82.4 minutes.

## Compatibility

- legacy `standard` migration timing remains at 144 blocks
- persisted timing-policy identifiers remain unchanged
- already-persisted schedule offsets remain unchanged
- anchor selection and retention for proofs built after the upgrade use the
  new age-4 cap
- Regtest and fast-Testnet delay parameters remain unchanged

## Checks

- `cargo fmt --all -- --check`
- `cargo test --lib` (475 passed, 1 ignored)


Thunnini:
@codex review

ValarDragon:
@codex review

ValarDragon:
My local codex is saying:

  - Low severity: ZIP318_ANCHOR_AGE_CAP is also used as a raw-block Regtest witness lookback in rust/src/wallet/sync/send.rs:2470. Reducing the cap to 4 shrinks that lookback from 16
    blocks to 4, while fast-Regtest anchor buckets are 12 blocks wide. Empty Regtest gaps could therefore make a valid witness appear unavailable. This does not affect Mainnet/Testnet,
    but the lookback should use a separate private constant—probably preserving 16 blocks.
    
  (TL;DR regtest issue)

PR #385, opened 2026-07-29
Defer migration anchor retention until NU6.3

ValarDragon:
## Summary

- keep normal sync batch sizes while scanning blocks at or before NU6.3
- split a batch exactly at the first post-NU6.3 block, then apply the existing
  96-block retention cap and checkpoint maintenance
- stop both the cap and maintenance when the final migration reference is
  released

## Why

An active migration currently caps every sync batch at 96 blocks and runs
migration-anchor retention after every batch. During a historical rescan this
also affects blocks before NU6.3, even though Vizor's migration anchor selection
already rejects boundaries at or before the NU6.3 activation height.

This makes a migration wallet rescan pre-NU6.3 history in many small batches and
repeatedly perform retention work that cannot discover an eligible migration
anchor.

The new boundary is derived from the network's NU6.3 activation height rather
than a mainnet literal. Existing retained checkpoints remain pinned during
pre-activation scanning; this change only defers repeated maintenance and the
small-batch policy until the scan reaches the first eligible height.

## Tests

- verifies normal batches and no retention through the activation block
- verifies an overlapping batch stops exactly at the activation boundary
- verifies post-activation batches retain the pruning-safe cap
- verifies releasing retention restores normal batches
- verifies the boundary comes from the network NU6.3 activation height
- verifies composition with the existing sandblasting batch policy
- `cd rust && cargo test` (480 passed, 1 ignored; Docker regtests remain ignored)

## API surface

No public or `pub(crate)` API changes.

### DIFF
--- a/rust/src/wallet/sync_engine/mod.rs
+++ b/rust/src/wallet/sync_engine/mod.rs
@@ -130,15 +130,35 @@ fn elapsed() -> String {
         .unwrap_or_default()
 }
 
-fn batch_size_for_range(base_batch_size: u32, start: BlockHeight, range_end: BlockHeight) -> u32 {
+fn batch_size_for_range(
+    base_batch_size: u32,
+    start: BlockHeight,
+    range_end: BlockHeight,
+    migration_anchor_retention_start: Option<BlockHeight>,
+) -> u32 {
     let start_u32 = u32::from(start);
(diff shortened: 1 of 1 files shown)


czarcas7ic:
Closing in favor of https://github.com/chainapsis/vizor-wallet/pull/390

PR #386, opened 2026-07-29
Defer transaction resubmission until scanning completes

ValarDragon:
## Summary

- defer startup resubmission while the wallet has pending scan ranges
- refresh the post-batch tip, recompute pending scan work, and only resubmit after that work is drained
- add regressions for both a newly discovered tip block and repeated 96-block historical batches

## Root cause

The resubmission candidate query relies on the wallet's local `mined_height`. During catch-up, a wallet-created transaction can already be mined on-chain while its local row still has `mined_height IS NULL` because the containing block has not been scanned yet.

Vizor previously ran resubmission at startup and after every scan batch. Migration anchor retention reduces historical batches to 96 blocks, so an already-mined transaction could be rebroadcast at every batch boundary until the scan reached its mined block.

## Impact

Resubmission now waits until there are no pending scan ranges. This gives scanning and enhancement the opportunity to record on-chain transactions as mined before the candidate query runs. Truly pending transactions remain eligible for automatic rebroadcast once the wallet is caught up.

No public or `pub(crate)` API surface changes.

## Validation

- `cargo test resubmit_`: 11 passed
- `cargo test`: 478 passed, 0 failed, 1 ignored; Docker regtest tests remain ignored as configured
- `cargo fmt`

### DIFF
--- a/rust/src/wallet/sync_engine/mod.rs
+++ b/rust/src/wallet/sync_engine/mod.rs
@@ -146,6 +146,12 @@ fn chain_tip_exclusive_end(current_tip_height: u64) -> BlockHeight {
     BlockHeight::from_u32(current_tip.saturating_add(1))
 }
 
+fn should_resubmit_after_tip_refresh(_previous_tip_height: u64, _fresh_tip_height: u32) -> bool {
+    // This preserves the current behavior while giving the regression test a
+    // narrow seam for the post-batch ordering decision.
+    true
+}
+
 fn scannable_batch_end(
     base_batch_size: u32,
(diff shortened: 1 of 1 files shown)


p0mvn:
Remade on the Vizor repo (with your commits preserved as author) plus a small completion-path hardening follow-up: https://github.com/chainapsis/vizor-wallet/pull/392

Please prefer #392; this draft can be closed once that lands.

p0mvn:
Closing as superseded by https://github.com/chainapsis/vizor-wallet/pull/392

PR #390, opened 2026-07-29
Continue using normal block batch over 96 block batching

czarcas7ic:
When migration anchor retention was introduced, Vizor was pinned to librustzcash `zcash_client_backend 0.24.0-rc.1` and `zcash_client_sqlite 0.22.0-rc.1`. In rc1, retention was recorded only after `insert_tree` had inserted and pruned a batch, so Vizor capped scans at 96 blocks—below the wallet's 100-checkpoint pruning window—to keep a needed migration anchor available for post-scan retention.

Vizor now uses rc4 of both crates. rc4 records retained checkpoint IDs before `insert_tree`, and ShardTree permits retaining an ID before that checkpoint exists. This applies the same ordering to Vizor's exact Orchard migration representatives, including empty logical boundaries, before each scan, so normal sync batch sizes are safe and the permanent 96-block cap is no longer needed.


czarcas7ic:
@codex review

PR #395, opened 2026-07-30
Keep locally created transaction lookups off lightwalletd

czarcas7ic:
This is the second change in the migration privacy stack.

When Vizor creates and submits a shielded transaction, it already has the complete raw bytes. While that transaction remains pending across newly scanned head blocks, librustzcash requests its status. Vizor previously answered by calling `GetTransaction` on the sync lightwalletd, unnecessarily revealing the txid. The recovery layer below this PR handles transactions once compact scanning restores a known mined height; this follow up covers the at-head interval before mining and the rare local enhancement request.

This change recognizes complete, expiring shielded transactions with local-origin evidence. It reprocesses stored bytes for `Enhancement`, and keeps `GetStatus` local until compact scanning mines the transaction or the fully scanned height reaches its expiry. Recovered, externally created, transparent-only, zero-expiry, and incomplete transactions retain the lightwalletd path.

Migration and PCZT fallback storage add the local-origin marker as best effort metadata. A marker failure is logged but cannot turn already successful transaction storage into a user-visible storage failure.

### DIFF
--- a/rust/src/wallet/sync/mod.rs
+++ b/rust/src/wallet/sync/mod.rs
@@ -31,6 +31,9 @@ mod proposal_locks;
 mod send;
 mod transactions;
 
+#[cfg(test)]
+pub(crate) use transactions::mark_transaction_created_locally;
+
 // Re-export the split submodules at the `wallet::sync` path so every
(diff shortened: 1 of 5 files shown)


czarcas7ic:
@codex review

PR #396, opened 2026-07-30
Route migration broadcasts through Zakura

czarcas7ic:
This is the third change in the migration privacy stack.

It routes newly created Ironwood preparation and Phase 2 broadcasts through a separate Zebra-compatible JSON-RPC relay for recognized built-in lightwalletd presets. Custom lightwalletd selections retain the existing submission path, and ordinary sends are unchanged. The submission target is saved per migration run, and relay responses, transaction IDs, expiry, shielding shape, redirects, payload size, and public-host HTTPS are validated. iOS background submission uses the same saved target.

Together with the preceding local lookup change, the sync lightwalletd receives neither the initial submission nor later transaction-specific status or enhancement lookups for these migration transactions.

This remains draft and blocked on production relay redundancy. Before merge, we need either additional independent Zakura endpoints or a mixed Zakura and lightwalletd submission pool that always selects a lightwalletd host different from the one used for sync. The managed mainnet default is currently `https://zakura-broadcast.valargroup.dev`; testnet, regtest, and Ironwood Masquerade can opt in through network-specific build defines.


czarcas7ic:
@codex review

czarcas7ic:
We cant merge this until we either get more Zakura endpoints from other people running, or we modify this PR to use both Zakura OR LWD endpoints for submission, and change it so that if we do use LWD for submission, its not the same host as the one we have set for sync.

PR #408, opened 2026-07-30
Drop raw blobs from the history bases query

p0mvn:
Stacked on #407 — review that one first; this PR's diff is only the second commit.

## Summary

`read_history_bases` selected from `v_transactions`. That view selects `transactions.raw` and aggregates over it, and SQLite cannot eliminate the column from inside the view's aggregate — so every reference materialized every raw transaction blob even though this path never reads them. The `account_uuid` predicate also could not be pushed through the view's `UNION` CTEs, so filtered and unfiltered counts cost the same.

This mirrors the view's aggregates as `HISTORY_BASES_CTE`, minus `raw` and the columns this module does not read, with the account filter applied before grouping. The `notes` and `sent_note_counts` CTEs are copied verbatim so their `UNION` de-duplication keeps identical row identity.

## Measurements

9-account wallet, 1008 txs, ~5714 Orchard notes, ~37 MB of raw blobs. macOS release build.

| | before #407 | after #407 | this PR |
|---|---|---|---|
| bases query (CPU, isolated) | ~1.8 s | ~1.8 s | **~0.04 s** |
| `get_transaction_history` | 2363–3732 ms | 1102–1198 ms | **46–436 ms (median ~112)** |
| account switch to rendered data | 2754 / 4472 ms | 1451 / 1911 ms | **335 / 374 / 444 / 793 / 976 / 1145 / 1363 ms** |

`getBalance` (205–1249 ms) is now the largest remaining cost and is the natural next target.

## Why the tests moved

The history fixtures define `v_transactions` as a **table** of hand-written aggregates, so a case can be stated directly — "this tx spent 1_265_000 and received 1_150_000". The real schema cannot express that in isolation:

- `notes` builds spent rows by joining `v_received_outputs` to `v_received_output_spends`, so a spent note must have been *received* by an earlier tx on the same account — which is then a history row too.
- `account_balance_delta` is `SUM(notes.value)`, structurally equal to `total_received - total_spent`. Three existing call sites declare values that violate this.
- A tx with no notes cannot appear at all, since the view is `FROM notes JOIN ...`.

25 of the 34 fixture-based tests spend, so rebuilding the fixtures on the real schema would have changed what each of them asserts — and those assertions are the specification of the display logic.

So the seam moved instead:

- `assemble_history` — the whole classification pipeline (summarize, suppress funding steps, classify, filter, sort, truncate) as a pure function over `TxBase` / `TxOutput`, no database.
- `history_from_fixture` — test entry point that feeds bases from the previous `v_transactions` query into the **real** `read_history_outputs` and the **real** `assemble_history`. Existing tests keep their fixtures and assertions verbatim; only the entry point changed, at 20 call sites.
- `history_bases_match_v_transactions` — `#[ignore]`d tripwire comparing `HISTORY_BASES_CTE` to the upstream view field-by-field on a librustzcash-built database.

## Tests

- `cargo test --lib`: **518 passed, 1 failed**. The failure, `expired_broadcasted_without_local_raw_stays_broadcasted_for_store_retry`, is pre-existing and reproduces identically on `main` at 319fc593 — verified in a clean worktree. This is parity with `main`.
- Equivalence tripwire against a 9-account mainnet wallet:

  ```
  VIZOR_HISTORY_EQUIV_DB=/path/to/zcash_wallet.db \
    cargo test --lib history_bases_match_v_transactions -- --ignored --nocapture
  → compared 1026 rows across 9 accounts ... ok
  ```

  Also confirmed it *fails* when the mirrored SQL is perturbed (`SUM(notes.spent_value) + 1`), naming the offending account and txid — so it is a real tripwire, not a vacuous pass.
- The same comparison was run independently through `sqlite3` across all 9 accounts before any Rust was written: md5-identical result sets, 1026 rows.

## Maintenance

`HISTORY_BASES_CTE` carries a permalink to the exact upstream definition it mirrors — `zcash_client_sqlite` 0.22.0-rc.4, `VIEW_TRANSACTIONS`, [`librustzcash@65a3add … db.rs#L1320-L1438`](https://github.com/zcash/librustzcash/blob/65a3add2f1d9b9ea455a71a9c33f9219dbc9e614/zcash_client_sqlite/src/wallet/db.rs#L1320-L1438). The vendored crate source was verified byte-identical to that revision. On a `zcash_client_sqlite` upgrade, run the tripwire and re-check the mirrored SQL.

Unrelated, noticed while here: `AGENTS.md` says `zcash_client_sqlite 0.20.2` but `Cargo.lock` pins `0.22.0-rc.4`.

### DIFF
--- a/rust/src/wallet/sync/transactions.rs
+++ b/rust/src/wallet/sync/transactions.rs
@@ -419,7 +419,7 @@ pub(crate) struct ExportBirthdayAnchor {
     pub block_height: u64,
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 struct TxBase {
     txid: Vec<u8>,
     transaction_id: i64,
@@ -575,6 +575,33 @@ pub fn get_transaction_history(
         &uuid_bytes,
         bases.iter().map(|base| base.txid.as_slice()),
     )?;
+    drop(read_tx);
+
+    Ok(assemble_history(
+        &bases,
+        &outputs_by_txid,
+        &uuid_bytes,
+        limit,
+    ))
+}
+
+/// Turn raw history rows into the display list.
+///
+/// This is the whole classification pipeline — summarize, suppress
+/// funding steps, classify, filter, sort, truncate — with no database
+/// access, so it can be exercised directly from `TxBase` / `TxOutput`
+/// values instead of through SQL fixtures. `read_history_bases` and
+/// `read_history_outputs` own the SQL side; their agreement with
+/// librustzcash's own schema is pinned by the regtest equivalence
+/// check rather than by synthetic in-memory tables, which cannot
+/// express states the real schema forbids (a spend, for instance,
+/// always implies a funding receive that is itself a history row).
+fn assemble_history(
+    bases: &[TxBase],
+    outputs_by_txid: &HashMap<Vec<u8>, Vec<TxOutput>>,
+    uuid_bytes: &[u8],
+    limit: Option<u32>,
+) -> Vec<TransactionInfo> {
     let summaries: HashMap<Vec<u8>, ActivitySummary> = bases
         .iter()
         .map(|base| {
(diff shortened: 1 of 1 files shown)


rust/src/wallet/sync/transactions.rs
```diff
@@ -2983,6 +3102,190 @@ mod tests {
...
+    ///
+    /// This is needed because the synthetic `v_transactions` table
+    /// doesn't have the note-derived aggregates that the real
+    /// `HISTORY_BASES_CTE` does. 
+    /// 
+    /// This helper allows us to test `read_history_bases` against the
+    /// synthetic `v_transactions` table.
+    /// 
+    /// The equivalency test checks that the two paths return the same rows.
+    fn read_history_bases_via_v_transactions(
```
p0mvn, line 3115: Note: this is a workaround to avoid a messier test rewrite. I investigated if there is a cleaner way, including reusing more of `librustzcash` APIs, but there weren't any

PR #424, opened 2026-07-31
Add mined_height to the store-retry test transactions fixture

czarcas7ic:
The store-retry migration test fails on current main because its `transactions` fixture models only `(txid, raw)`, while expiry counting now runs `local_denomination_chain_identity` per row, whose test-only fallback queries `transactions.mined_height`. Adds the nullable column; the row stays unmined, so the store-retry semantics under test are unchanged. Verify with `cd rust && cargo test --lib expired_broadcasted_without_local_raw`.


czarcas7ic:
Folding this into the migration schedule PR as its first commit; a separate PR for a four-line test fixture is more overhead than it is worth.

PR #433, opened 2026-07-31
perf(desktop): trigger a migration advance when a transfer comes due

p0mvn:
First of a desktop-focused perf stack. Each PR is self-contained and measured.

## The defect

`_shouldAdvance` bypasses its rate limit whenever `_advanceProgressKey` changes. That key is built from run id, phase, and per-part counts — **it contains no height**. Crossing a scheduled broadcast height changes none of those fields, so a transfer becoming due is invisible to the trigger.

The broadcast therefore waits out `_migrationAdvanceInterval` (30s), sampled at `_migrationStatusPollInterval` (15s), before it is even attempted.

This is desktop's timed-submit path: `_shouldAdvance` returns true by phase for `broadcast_scheduled` / `broadcasting` / `waiting_confirmations`, then `_advance` → `broadcastOneDueOrchardMigrationTransaction`.

## Measured

Delay from "transfer becomes due" to "broadcast attempted", over one full advance cycle (300k samples, `poll=15s`, `interval=30s`):

| | delay |
|---|---|
| min | 0.00s |
| mean | **15.00s** |
| median | 15.00s |
| p90 | 27.00s |
| max | **30.00s** |

Against ~75s Zcash blocks that is **20% of a block on average, 40% worst case** — enough to routinely push a transfer into the block after its scheduled height.

The defect is also demonstrated directly against the real coordinator. Before this change, `test/features/migration/ironwood_migration_desktop_advance_trigger_test.dart` showed:

- crossing the scheduled height → **no advance** (broadcast count stayed at 1)
- changing `broadcastedTxCount`, a field that *is* in the key → advance fired immediately

which isolates the cause: the trigger mechanism works, height just is not part of it.

## The change

Add `_dueScheduledBroadcastCount(status)` to `_advanceProgressKey` — the number of scheduled transfers whose height the wallet has already observed.

Counting rather than flagging means a **second** transfer coming due is also a key change. The count is stable between crossings, so an advance that cannot broadcast yet falls back to the ordinary 30s interval instead of re-firing on every poll.

Height comes from `_observedBroadcastHeight()` (`min(scanned, tip)`), matching the existing `_hasDueScheduledBroadcast`. That is deliberately conservative: if sync lags the tip the trigger waits, and the interval fallback still covers it. Rust independently gates on `chain_tip_height`, so this only ever advances the moment, never the decision.

## Scope

`_advanceProgressKey` is shared, so mobile gets the same immediate trigger. Mobile advances are additionally gated on `foregroundProgressPermits`, so this changes *when* a permitted advance runs, not whether one may.

## Testing

- New `ironwood_migration_desktop_advance_trigger_test.dart` (desktop lane, 4 tests): crossing advances immediately; a still-due transfer does **not** hot-loop; a second transfer coming due fires again; the pre-existing progress-key mechanism still works.
- Desktop lane `test/features/migration/ test/providers/`: **525 passed**, 3 skipped.
- Mobile lane, same paths: `196 passed, 15 failed` — **identical to this branch's base**, verified by re-running with the change stashed. Pre-existing, confirmed by the author.
- `flutter analyze` on both changed files: no issues.

### DIFF
--- a/lib/src/features/migration/providers/ironwood_migration_coordinator_provider.dart
+++ b/lib/src/features/migration/providers/ironwood_migration_coordinator_provider.dart
@@ -1307,12 +1307,36 @@ class IronwoodMigrationCoordinator
       status.broadcastedTxCount,
       status.confirmedTxCount,
       status.signedChildPcztCount,
+      // Crossing a scheduled height is what makes a transfer broadcastable,
+      // but it changes nothing else in the status. Without it here the
+      // transfer waits out `_migrationAdvanceInterval` before an advance even
+      // attempts it (mean 15s, up to 30s at the 15s poll cadence).
+      _dueScheduledBroadcastCount(status),
       for (final part in status.parts) ...[
         part.partIndex,
         part.state.name,
         part.confirmationCount,
       ],
     ].join(':');
+  }
+
+  /// Scheduled transfers whose target height the wallet has already observed.
+  ///
+  /// Counted rather than flagged so that a second transfer coming due is also
+  /// a key change. The count is stable between crossings, so an advance that
+  /// cannot broadcast yet still falls back to the ordinary interval instead of
+  /// re-firing on every poll.
+  int _dueScheduledBroadcastCount(rust_sync.MigrationStatus status) {
+    final currentHeight = _observedBroadcastHeight();
+    if (currentHeight <= 0) return 0;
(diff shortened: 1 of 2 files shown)


p0mvn:
Superseded by #438 — same change, rebased standalone onto `perf-summary-main-2`.

PR #434, opened 2026-08-01
perf(desktop): return from a migration stop once the abandon commits

p0mvn:
Stacked on #433.

## The defect

`_runStop` did four things in sequence, and the schedule screen awaits all of them before it navigates (`schedule.dart:182` — `await coordinator.stop(...)` then `context.go('/home')`):

1. `await` any in-flight advance — **required**, the Rust `ActiveIronwoodMigration` guard rejects a concurrent abandon
2. `await service.stop(...)` — the durable abandon, **required**
3. `await refreshAfterSend()` — balance + history
4. `await refreshNow()` — full coordinator status sweep

Steps 3 and 4 run **after the durable stop has committed**. They cannot change the outcome, and the user is leaving the screen they refresh.

This is the same shape as #416 in the #420 series: the expensive work wasn't what made it feel slow, the blank/spinner window was.

## Measured

Fakes sized from the #420/#429 figures on a 10-account mainnet wallet — 150ms balance refresh, 200ms status read, 100ms durable abandon:

| | `stop()` blocks for |
|---|---|
| before | **666 ms** |
| after | **312 ms** |
| | **−354 ms (−53%)** |

The 312ms that remains is the status read inside `service.stop` plus the abandon itself. Both legitimately precede the return.

## A latent bug this also fixes

`await refreshNow()` sat inside the outer `try`. A failure there was caught by the handler, which set the account error and **rethrew** — so a stop that had durably succeeded surfaced to the user as a failure, inviting them to retry a destructive action that had already happened.

`refreshAfterSend` was already guarded against exactly this ("must not invite the user to repeat the destructive action"); the status sweep was not. Both are now logged instead.

## The change

Move steps 3 and 4 into `_refreshAfterStop()` and `unawaited(...)` it. Deferred, not dropped — both still run, both re-check `ref.mounted`, both log their own failures.

The in-flight advance await in step 1 is deliberately untouched: dropping it would let an advance broadcast while the abandon reconciles, and Rust's `ActiveIronwoodMigration::acquire` would reject the abandon outright with "already running".

## Scope

`stop()` is shared, so mobile gets the same deferral. Mobile stop screens navigate on the same future.

## Testing

- New `ironwood_migration_desktop_stop_latency_test.dart` (desktop lane, 4 tests): the timing bound above; the deferred refreshes still run; a genuine stop failure still throws and still lands in `state.errors`; the stopping lease is cleared before returning.
- Desktop lane `test/features/migration/ test/providers/`: **529 passed**, 3 skipped.
- Mobile lane, same paths: `196 passed, 15 failed` — identical to base, pre-existing.
- `flutter analyze` on changed files: no issues.

### DIFF
--- a/lib/src/features/migration/providers/ironwood_migration_coordinator_provider.dart
+++ b/lib/src/features/migration/providers/ironwood_migration_coordinator_provider.dart
@@ -625,15 +625,10 @@ class IronwoodMigrationCoordinator
       state = state.copyWith(
         errors: Map<String, String>.from(state.errors)..remove(accountUuid),
       );
-      try {
-        await ref.read(syncProvider.notifier).refreshAfterSend();
-      } catch (error) {
-        // The durable stop has already committed. A balance refresh failure
-        // must not invite the user to repeat the destructive action.
-        log('Ironwood migration post-stop balance refresh failed: $error');
-      }
-      if (!ref.mounted) return;
-      await refreshNow();
+      // The durable stop has committed. The schedule screen navigates on this
+      // future, so anything awaited past this point is pure spinner: the
+      // balance refresh and status sweep measured 350ms of a 666ms stop.
+      unawaited(_refreshAfterStop());
     } catch (error) {
       if (ref.mounted) {
         state = state.copyWith(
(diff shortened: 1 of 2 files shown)


p0mvn:
Superseded by #439 — same change, rebased standalone onto `perf-summary-main-2` instead of stacked on #433.

PR #435, opened 2026-08-01
fix(desktop): let user-driven migration writes outlast a sync batch

p0mvn:
Stacked on #434.

## Correcting two premises

This PR began as "chunk the scan under the global write lock." Measuring first killed that: only **2 of 69** migration DB opens take `with_wallet_db_write_lock`, so migration work does not queue behind the process mutex at all — it contends at the SQLite level.

Then testing the *real code shape* corrected the replacement too. Raising the busy timeout alone would have fixed only part of the problem. Details below.

## The defect

`wallet/db.rs` has a deliberate timeout hierarchy:

| constant | value | intent |
|---|---|---|
| `WALLET_DB_BUSY_TIMEOUT` | 10s | "User-driven wallet operations can afford a longer wait for a short sync write." |
| `SYNC_DB_BUSY_TIMEOUT` | 2s | must not stretch cancel |
| `READ_DB_BUSY_TIMEOUT` | 2s | reads |

Migration writes opened with the 2s **read** timeout — including exactly the reported paths: `abandon_run` (stop), `mark_pending_broadcasted` / `mark_pending_broadcast_attempted` (scheduled broadcast), `replace_resigned_pending_parts` / `reschedule_overdue_pending_txs_with_options` (recover).

`scan_cached_blocks` runs a whole batch in one transaction (desktop `BATCH_SIZE_FOREGROUND = 2000`), so the writer is held for seconds on a large wallet. An overlapping migration write exhausts 2s and gets `SQLITE_BUSY` — an **error, not a delay**, which fits "when there's an issue, recover button" far better than slowness does.

## Why the timeout alone was not enough

These functions begin with `unchecked_transaction()`, which is **DEFERRED**. A deferred transaction that reads before it writes must upgrade its read lock, and SQLite **deliberately skips the busy handler for that upgrade** — two connections both waiting to upgrade would deadlock — so it returns `SQLITE_BUSY` instantly regardless of the timeout.

Probed directly against a real WAL database with a writer held longer than the timeout:

| transaction shape | 10s timeout |
|---|---|
| write-first deferred | **Ok** |
| read-then-write deferred | **Err(DatabaseBusy)**, immediately |

Of the 23 functions, **19 are write-first** (helped by the timeout) and **4 read first** — `replace_resigned_pending_parts`, `reconcile_denomination_stage_chain_state_with_rng`, `apply_accepted_migration_outbox_receipt`, `reset_migration_children_for_reorged_denominations`. Two of those are on the recover path. For them the timeout change would have been inert.

## The change

`open_wallet_write_conn` sets **both**:

- `WALLET_DB_BUSY_TIMEOUT`, so the write waits out a sync batch;
- `TransactionBehavior::Immediate`, so the wait can happen at all — the write lock is taken at `BEGIN`, so there is no upgrade for SQLite to bail out of.

`unchecked_transaction()` reads that connection-level default (rusqlite 0.37 `set_transaction_behavior`), so no call site needed restructuring and no borrow-checker churn. 27 migration write sites now open through it. Three functions already used `TransactionBehavior::Immediate` explicitly, which is what suggested the mechanism.

Deliberately excluded: `stage_migration_anchor_retention_references` and `finish_migration_anchor_retention_releases`. The sync engine calls them from inside `with_wallet_db_write_lock("sync_engine.retain_and_scan_blocks")` (`send.rs:2995`, `send.rs:3014`), where a 10s wait would stretch the scan and the cancel check after it.

## Measured

Four tests in `wallet::db::tests`, each against a real WAL file with a background thread holding `BEGIN IMMEDIATE` for `2 x READ_DB_BUSY_TIMEOUT`:

| test | asserts |
|---|---|
| `read_timeout_write_fails_under_a_long_sync_write` | 2s write → `SQLITE_BUSY` |
| `wallet_timeout_write_survives_a_long_sync_write` | 10s write → succeeds |
| `deferred_read_then_write_ignores_the_busy_timeout` | deferred read-then-write → fails despite 10s |
| `immediate_read_then_write_waits_out_the_sync_write` | same transaction under `Immediate` → succeeds |

The hold is derived from the constant under test rather than hardcoded, and the busy assertion matches `SQLITE_BUSY` specifically rather than any error.

## Trade-off

`Immediate` takes the write lock at `BEGIN` rather than at first write, so these transactions hold it marginally longer. They are short migration-table writes, and the alternative is failing outright on a shape where the timeout cannot help. Reviewers should sanity-check that none of the 27 holds a transaction open across expensive work.

## What this does not do

Scan chunking is **deferred, not dismissed**. Shortening the writer hold helps independently. But each sub-chunk needs a `ChainState` for its start height (a `get_tree_state` round trip), it sits on the reorg/continuity recovery path, and the benefit cannot be quantified without a real multi-account wallet or a regtest run.

Related: the uncommitted working-tree change raising desktop `BATCH_SIZE_SANDBLASTING` to 2000 pushes the other way — fewer round trips, longer writer holds. This PR makes that trade safer either way.

## Testing

- `cargo test --lib`: **540 passed, 0 failed**, 3 ignored (536 on base + 4 new).
- No behaviour change without contention: a busy timeout is an upper bound on waiting, not a delay.

### DIFF
--- a/rust/src/wallet/db.rs
+++ b/rust/src/wallet/db.rs
@@ -14,11 +14,19 @@ use crate::wallet::network::WalletNetwork;
 pub(crate) type WalletDatabase = WalletDb<rusqlite::Connection, WalletNetwork, SystemClock, OsRng>;
 
 /// User-driven wallet operations can afford a longer wait for a short sync write.
+///
+/// This is the tier for any write a user is waiting on, including migration
+/// stop, scheduled broadcast, and coordinator advance. `scan_cached_blocks`
+/// runs a whole batch inside one transaction, so on a large wallet the writer
+/// can be held for seconds; a user-driven write must outlast that rather than
+/// fail. The two `wallet::db::tests` busy-timeout tests pin both directions.
 pub(crate) const WALLET_DB_BUSY_TIMEOUT: Duration = Duration::from_secs(10);
 /// Account creation/import runs after sync is paused, so a shorter wait exposes real stalls.
 pub(crate) const ACCOUNT_MUTATION_DB_BUSY_TIMEOUT: Duration = Duration::from_secs(5);
 /// The sync loop should absorb brief read/write overlap without stretching cancel too far.
 pub(crate) const SYNC_DB_BUSY_TIMEOUT: Duration = Duration::from_secs(2);
+/// Reads only. A write opened with this gives up while a sync batch is still
+/// committing -- use [`WALLET_DB_BUSY_TIMEOUT`] for writes a user is waiting on.
 pub(crate) const READ_DB_BUSY_TIMEOUT: Duration = Duration::from_secs(2);
 
 /// Seqlock-style write epoch for in-process wallet-summary cache invalidation.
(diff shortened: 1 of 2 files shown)


p0mvn:
Dropping this one.

It does not address the reported symptom. The reports were of **slowness**; this converts a `SQLITE_BUSY` **failure** into a wait, so in the contended case a single attempt gets *longer*, not shorter. It only explains a slow report through a fail -> "Try again" -> retry cycle, and if migration writes were always completing inside 2s it changes nothing at all.

What the tests here proved is that the failure mode is *possible*, not that it *occurs*. That premise is unmeasured and it is the entire value of the change.

Cheap way to revive it: the SQLite message propagates into the error string (`abandon_run` wraps it as `Mark migration stopped: {e}`, and the coordinator stores `error.toString()` in `state.errors` and logs it), so affected users' logs and Sentry events would contain `database is locked`. If that string shows up, this is worth reopening as a correctness fix. The `TransactionBehavior::Immediate` finding is the part worth keeping — a long busy timeout alone is inert for a deferred read-then-write transaction, which covered 4 of the 23 functions including two on the recover path.

PR #436, opened 2026-08-01
perf: derive the migration payload key once per read, not once per row

p0mvn:
Stacked on #435. Final PR of the desktop stack.

## The defect

`secret_payload::derive_key` runs `KDF_ITERATIONS = 100_000` PBKDF2-HMAC-SHA256 iterations. Measured at **~25ms per derivation** in release on an M-series Mac — the *fastest* hardware this ships on.

Eight migration functions decrypt or encrypt one payload per row and called it **inside the loop**, re-deriving the same key from the same password and salt every iteration:

| function | derivations per row |
|---|---|
| `denomination_stages_for_run` | **3** |
| `signed_child_pczts_for_run` | 2 |
| `insert_signed_child_pczts_with_tx` | 2 |
| `export_scheduled_migration_outbox` | 1 |
| `broadcasted_pending_txs_missing_local_identity` | 1 |
| `insert_pending_txs_with_tx` | 1 |
| `replace_resigned_pending_parts` | 1 |
| `pending_raw_denomination_stages` | 1 |

These are on the paths in the bug reports — outbox export and `broadcasted_pending_txs_missing_local_identity` run on every broadcast pass, so the cost recurs per poll, not once per run.

## Measured

`kdf_reuse_benchmark`, 20 rows x 3 payloads — the shape of a denomination-stage read:

```
kdf: 20 rows x 3 payloads (60 derivations -> 1)
  per-call derivation : 1.488214167s
  derived once        :   25.127792ms
  saved               :  1.463086375s (59.2x)
```

Run it with `cargo test --release --lib kdf_reuse -- --ignored --nocapture`. It is `#[ignore]`d so a timing test never gates CI.

## The change

Add `PayloadKey`: derive once, then `encrypt`/`decrypt` against the held key. `encrypt_payload` and `decrypt_payload` become one-line delegates, so **every existing call site is untouched** — only the eight loops opt in.

**Deliberately not a process-wide key cache.** A global would keep derived key material resident across unrelated operations. `PayloadKey` is scoped to the call that builds it, so key material lives no longer than the password the caller already holds. It stays in `Zeroizing`.

## Correctness

Two always-on tests, because a loop conversion must not change what it reads or writes:

- `payload_key_matches_the_free_functions` — round-trips **both** directions: written by `PayloadKey` / read by the free function, and vice versa. Same ciphertext format, same key.
- `payload_key_rejects_a_wrong_password` — a wrong password still fails to decrypt.

## Testing

- `cargo test --lib`: **540 passed, 0 failed**, 4 ignored.
- Run 5x consecutively to check for flakes: 5/5 clean. (One transient failure was seen once mid-session while a release build ran concurrently, and did not reproduce; the two timing-sensitive tests added in #435 were hardened in that PR — the busy-timeout hold is now derived from the constant under test and the assertion matches `SQLITE_BUSY` specifically rather than any error.)
- `cargo fmt --check` reports only the 12 pre-existing drift sites in `migration/tests.rs`, which are present on the base branch and deliberately left alone.

### DIFF
--- a/rust/src/wallet/secret_payload.rs
+++ b/rust/src/wallet/secret_payload.rs
@@ -64,15 +64,53 @@ pub fn derive_password_verifier_base64(
     Ok(STANDARD.encode(key.as_slice()))
 }
 
+/// A payload key derived once for one `(password, salt)` pair.
+///
+/// [`derive_key`] runs `KDF_ITERATIONS` PBKDF2-HMAC-SHA256 iterations, which
+/// measures ~27ms per derivation on an M-series Mac in release. Loops handling
+/// one payload per row re-derived the same key on every iteration; build this
+/// once outside the loop instead.
+///
+/// Deliberately not a process-wide cache. The key lives exactly as long as the
+/// call that needs it, so this does not widen the window in which key material
+/// is resident relative to the password the caller already holds.
+pub struct PayloadKey {
+    key: Zeroizing<[u8; KEY_LEN]>,
+}
+
+impl PayloadKey {
+    pub fn new(password: &[u8], salt: &[u8]) -> Self {
+        Self {
+            key: derive_key(password, salt),
+        }
+    }
+
+    fn cipher(&self) -> Result<Aes256Gcm, String> {
+        Aes256Gcm::new_from_slice(self.key.as_slice())
+            .map_err(|e| format!("Failed to initialize AES-GCM: {e}"))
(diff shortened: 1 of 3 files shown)


p0mvn:
Superseded by #437, which is the same change rebased standalone onto `perf-summary-main-2` instead of stacked behind #435 (dropped).

PR #453, opened 2026-08-03
Bound Keystone migration signing rounds

ValarDragon:
## Summary

- cap one Keystone migration request at 35 transactions, below the firmware's 40-message protocol ceiling
- compute each compact PCZT's standalone size after applying the same `resolve_fields()` normalization as Keystone
- partition QR signing rounds by both the 35-transaction cap and a 490 KiB normalized-byte budget
- carry overflow transactions into subsequent signing rounds instead of dropping them

## Root cause

Keystone applies its 512 KiB limit twice: first to the incoming compact request, and again after validating each PCZT, restoring compact fields, and rebuilding the checked batch.

Migration denomination transactions contain 16 Orchard actions with empty memos. Compact redaction replaces each 580-byte encrypted ciphertext with an empty memo representation and omits the 32-byte `cv_net` and `cmx` fields. The production-padding fixture therefore measures 7,113 bytes on the QR wire but 17,423 bytes after Keystone normalization.

Thirty such transactions total 522,690 normalized bytes before the outer envelope, which explains the reported failure near the firmware limit even though the compact request is much smaller. A percentage discount on compact bytes would not be reliable because the expansion depends on PCZT contents.

## Impact

The representative 30-transaction case is split into rounds of 28 and 2. Smaller transactions can still use up to 35 messages in one round. Every transaction remains in the signing session.

## Validation

- `flutter_rust_bridge_codegen generate`
- `cargo test padded_denomination_split_builds_exactly_sixteen_actions --lib`
- `cargo test batch_signer_redaction_compacts_and_preserves_signable_spends --lib`
- `cargo test migration_batch_signing_selector_uses_the_product_message_limit --lib`
- `fvm flutter test test/features/migration/ironwood_migration_flow_screen_test.dart` (109 tests)
- `fvm flutter analyze` (23 pre-existing unused-code warnings; no new analyzer issues)

### DIFF
--- a/lib/src/features/migration/screens/ironwood_migration_flow/keystone_signing.dart
+++ b/lib/src/features/migration/screens/ironwood_migration_flow/keystone_signing.dart
@@ -1810,14 +1810,12 @@ List<rust_sync.KeystoneSignedMigrationMessage> _signedMigrationMessagesFor(
 // The Keystone firmware enforces two independent caps on one signing round:
 // the message count (`signingBatchLimit`) and a 512 KiB ceiling that covers
 // both the canonical PCZT byte total and the request-id + Postcard envelope
-// (`ZCASH_SIGN_BATCH_MAX_TOTAL_BYTES` in rust/src/wallet/keystone.rs). Rounds
-// must stay under both, or QR encoding rejects the round after the user has
-// already approved the migration.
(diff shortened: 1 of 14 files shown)


czarcas7ic:
Replaced by https://github.com/chainapsis/vizor-wallet/pull/459

PR #458, opened 2026-08-04
Rebase migration schedule onto finalization height

p0mvn:
## Summary

- durably rebase the initial unpromoted signed-child schedule when proof finalization first becomes possible
- preserve approved per-child offsets while preventing the anchor-readiness wait from making transfers immediately overdue
- retain existing catch-up and re-sign behavior when rebasing would cross an expiry bucket or cannot safely preserve the approved origin
- project finalized child timing from persisted heights and surface scheduled transfers ahead of later proof windows

## Motivation / Why

Signed migration schedules are created before denomination outputs become provable. By the time proof finalization can run, several approved transfer heights may already be overdue, causing catch-up scheduling to replace the intended tapered spacing. Anchoring the initial schedule to the actual finalization height preserves those approved offsets without changing later recovery behavior.

Observed run:

```
schedule origin:    3435169
denomination mined: 3435172
proof-ready:       ~3435409   (240-block readiness delay)
approved offsets:   0, 23, 81, 183, 311
overdue at ready:   0, 23, 81, 183
```

Starting the approved offsets at the height where finalization actually runs removes the routine trigger. Anchoring to the finalization height rather than the chain-derived proof-ready height also means the origin is not predictable from the denomination transaction, and it stays correct when the app is opened well after readiness.

## Human Context

Rebase migration schedule onto proof finalization height

We observed that there were surprisingly short clusters at the beginning of the schedule in manual testing on mainnet. What happened: the system was defining the schedule relative to the height of schedule creation without accounting for anchor delays.

By the time their anchor became eligible, many notes would become overdue and get rescheduled. Rescheduling deliberately had shorter mean, resulting in visible clusters in the beginning.

This PR "rebases" the schedule from the original height against the heights proofs can be constructed, eliminating the inevitable recreation and clustering.

Tested on both Regtest and Mainnet. Several rounds of AI reviews passed with no outstanding blockers.

## Tests

- `cargo test --lib wallet::sync::migration::tests` — 154 passed
- `fvm flutter test test/features/migration/ironwood_migration_presentation_test.dart` — 36 passed
- `fvm flutter analyze lib/src/features/migration/models/ironwood_migration_presentation.dart test/features/migration/ironwood_migration_presentation_test.dart`
- Rust formatting and `git diff --check`

### DIFF
--- a/lib/src/features/migration/models/ironwood_migration_presentation.dart
+++ b/lib/src/features/migration/models/ironwood_migration_presentation.dart
@@ -292,8 +292,7 @@ MigrationNextActionPresentation migrationNextActionPresentation({
   final proofWindowIsNext =
       nextProofWindowHeight != null &&
       proofWindowPartIndices.isNotEmpty &&
-      (earliestScheduledHeight == null ||
-          nextProofWindowHeight < earliestScheduledHeight);
+      earliestScheduledHeight == null;
   if (proofWindowIsNext) {
     final windowAmount = parts
         .where((part) => proofWindowPartIndices.contains(part.partIndex))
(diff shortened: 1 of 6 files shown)


Thunnini:
@codex review

PR #480, opened 2026-08-07
Enable Tor on mobile

piatoss3612:
Rebuild of #473, closed to regroup. Same scope, reorganized as eight logical commits built from that branch's final reviewed state — plus fixes for the three findings that were still open when it closed. The old branch `rowan/mobile-tor` remains as the review archive; the background lane and Rust core here are byte-identical to it.

## Commits

1. **Persist the Tor route on the strict side** (desktop) — the saved route may be stricter than the enforced one, never laxer. Enabling writes before anything changes, so a failed save aborts a toggle that never happened; disabling writes after the switch, with the direct-request gate following the runtime. This replaces the closed PR's ordering and fixes its open P1: no state exists anymore in which a relaunch or cold wake reads direct against a session that was told Tor.
2. **Cancellable pending connection** (desktop) — the toggle is an escape from the fail-closed connecting window instead of disabling itself for exactly that window.
3. **Refuse a route nothing has declared** (rust) — undeclared ≠ direct at the one chokepoint every lightwalletd connection passes; adoption of a persisted route is atomic with the state change; a bootstrap stops when its route goes away; the deadline covers the init-lock wait.
4. **Dormancy and the background wake lease** (rust) — idle clients sleep with the app lifecycle; a pass masks the intent with a self-lapsing, request-renewed lease; a cold background process adopts the intent it was never given.
5. **Enable Tor on iOS and Android** — startup gate removed, sandbox-aware fs-mistrust, guard state kept out of backups and device transfer, a launch-owned dormancy lifecycle that exists even on locked launches, and a hand-run live-Tor probe tagged out of the automated lanes.
6. **iOS background lane** — route declaration before any native call, charger+unmetered affordability, per-request-type gates, the wallet-mutation fence kept off the uncancellable bring-up, and a main-thread power monitor whose every read requests a fresh sample (fixes the open staleness P2: a suspended process no longer answers with the charger state from before the suspension).
7. **Mobile Tor control and route-truthful copy** — every failure state names the route actually carrying requests; migration background copy reads the saved route via a new `savedRouteTorEnabled` (fixes the open P2: the native lanes read the saved preference, and the two diverge in exactly one state).
8. **Docs** — the declaration contract and both saved-route rules in AGENTS.md.

## Verification

- Rust: 634 passed / 0 failed (every commit in the series compiles and passes independently for the Rust stages).
- Desktop lane: 2250 passed / 30 failed — the pre-existing main set.
- Mobile lane: 793 passed / 18 failed — 17 pre-existing plus one (`mobile_onboarding_progress_test`) confirmed failing on clean origin/main.
- iOS simulator: full app build; RunnerTests 223 passed, including new coverage for the power-staleness fix; live Tor bootstrap probe passes on the simulator.
- New behavioural tests were each verified to fail with only the production change reverted (toggle ordering ×2, saved-route migration copy, route chokepoint, claim atomicity, bootstrap abandon, dormancy default).

Still device-only: whether the fs-mistrust exemption is required on real hardware (recorded in the probe), power/network state transitions on device, and cancellation during a genuine cold bootstrap.

https://claude.ai/code/session_01R9nkJCPTCahMjGiDgqFNrV

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -545,10 +545,351 @@ Swift BackgroundMigrationPreparationManager
   `com.keplr.vizor.sync` is cancelled once at launch as a tombstone for requests
   submitted by older builds; no handler is registered for it.
 
+### Declaring the network route
+
+The desired route lives in Rust process memory, and a background wake can be a
+cold launch where the Dart layer never ran. So "nobody has declared yet" is a
+distinct state from "the user chose direct", and Rust refuses it at the route
+chokepoint rather than treating it as direct.
+
+- **Every native entry point that may reach lightwalletd declares the route
+  before its first network call**, by reading the saved preference and calling
+  `zcash_network_privacy_mark_tor_desired` or
+  `zcash_network_privacy_mark_direct_route`. Declaring samples nothing, never
+  bootstraps, and is ignored once the process has decided its own route, so it
+  is safe to do first and unconditionally.
+- A lane that forgets fails closed on its first connection with
+  `Network route has not been declared for this process`. That is the intended
+  outcome: the alternative is reaching lightwalletd over clearnet on a
+  Tor-configured wallet, silently and only in the lane that forgot.
+- Adding a background entry point therefore means adding a declaration. Nothing
+  fails to compile without one, and the Rust-side test that pins this
+  (`an_undeclared_route_is_refused_instead_of_resolving_to_a_direct_connection`)
+  covers the chokepoint, not your new caller.
+- **On a Tor route, background passes do no network work at all.** A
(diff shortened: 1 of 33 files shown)


piatoss3612:
Closing in favor of the reduced series on `rowan/mobile-tor-lite`: all background Tor machinery is removed per review economics — the background migration managers are now byte-identical to main, and a single launch-time route declaration plus the Rust chokepoint keeps a Tor wallet's cold background wake fail-closed. Replacement PR follows.

https://claude.ai/code/session_01R9nkJCPTCahMjGiDgqFNrV

PR #495, opened 2026-08-10
[mobile] Fix duplicate recipient selection in Pay and Swap

piatoss3612:
## Problem
Mobile Pay had the same address-only recipient ambiguity and quote-fetch timing issue as desktop. The selected duplicate contact also needed to survive route handoff into review and later Activity rendering.

## Solution
- Pass the explicit recipient selection through the mobile review route.
- Keep contact row taps on the recipient step and fetch the quote only from Continue.
- Preserve the selected contact ID through mobile Pay, Swap review, and Activity.
- Show a bare check only for duplicate-address contact groups with a fixed trailing slot.
- Gate recipient actions on the initial address-book load and use the app loader while fetching a quote.

## Stack
3 of 3. Depends on #494, which depends on #493.

## Verification
- fvm flutter test mobile lane: 46 focused tests
- fvm flutter analyze: 12 changed mobile files
- git diff --check
- Deterministic mobile recipient-step capture verified

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a></sub>

### DIFF
--- a/lib/src/core/navigation/mobile_routes.dart
+++ b/lib/src/core/navigation/mobile_routes.dart
@@ -19,6 +19,7 @@ import '../../features/migration/screens/ironwood_migration_flow_screen.dart'
         MobileIronwoodMigrationKeystoneDenominationSignScreen;
 import '../../features/pay/screens/mobile/mobile_pay_screen.dart';
 import '../../features/pay/screens/mobile/mobile_pay_submitted_screen.dart';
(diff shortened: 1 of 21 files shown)


piatoss3612:
Folded into #494 so the desktop and mobile UI changes stay together in the existing rowan/vzr-136-desktop branch and PR.

PR #506, opened 2026-08-13
Support negotiated PIR layout and round-bound vote shares

czarcas7ic:
Updates Vizor to the published `zcash_voting 3.0.0` release and regenerates the Rust/Dart bridge for two voting API changes: `PirLayout.polyLen` is carried from dynamic config into PIR work, and `VoteShareWire.voteRoundId` is carried from Rust so live and recovered share JSON can be forwarded without wallet-side injection. It also moves the related released crate family to `pir-types 0.3.0`, `pir-client 0.4.0`, `voting-circuits 0.10.0`, `vote-commitment-tree 0.5.0`, and `vote-commitment-tree-client 0.7.0`.

The stable release centralizes complete 16-share initial-target spreading. When multiple helpers are configured, no helper is selected for every share. Vizor already passes the complete batch to `plan_share_submissions`, so no additional wallet routing, persistence, or generated API change is needed for that policy.

Deployment remains gated on each target environment serving the negotiated polynomial length and using auth-v2-signed round entries. Do not ship this wallet against an environment before both changes are live.


czarcas7ic:
@codex review

PR #508, opened 2026-08-13
DRAFT: Patch Vizor to Zakura libraries

czarcas7ic:
Builds on the stable `zcash_voting 3.0.0` family from its base while applying a root `[patch.crates-io]` for the complete related crate family at Zakura revision `7490b608930de93ee69ade037c7aea4bf8b29f32`. The lockfile converges all 13 patched crates on that source, including `halo2_proofs 0.3.5` and `orchard 0.15.5`.

This is a compatibility experiment for using the optimized Zakura implementations as drop-in replacements without changing Vizor's dependency declarations. The final tree passes `cargo metadata --locked`, targeted inverse dependency trees, `cargo build --release --locked --lib`, `cargo test --locked` (621 passed, 4 ignored), and `git diff --check`.


ValarDragon:
Usefully this speeds up all your proving times, and will improve your trial decryption time by 15% on an iPhone

PR #512, opened 2026-08-15
Reduce startup and sync latency

ValarDragon:
## Summary

- Cache wallet support-directory and randomized DB-name lookups, invalidate the DB-name cache safely on reset, and start independent bootstrap storage/Rust reads together.
- Pipeline sync network and database work: concurrent Sapling/Orchard/Ironwood subtree downloads, transparent UTXO streams, compact-block/tree-state fetches, and next-batch prefetch.
- Avoid the per-batch resubmit cost when no raw pending transaction exists by using a prepared base-table preflight on a persistent read-only connection.
- Reconcile fresh full tip identities before completion or resubmission, prove lower-tip reorgs before rewinding, require exact DB/network completion, and reject incomplete or malformed lightwalletd responses.

## Expected impact

The main user-visible wins should be a shorter cold-start wait and substantially less wall-clock time for long import or recovery scans. A wallet already near the chain tip has only a few batches to process, so its absolute speedup should be smaller. This PR does not claim an end-to-end benchmark result yet; the expectations below come from removed serialized waits and operation counts.

- **Startup:** bootstrap completes before `runApp()`, so its latency directly affects how long the user remains on the native splash or blank launch state. Independent secure-storage reads and Rust snapshot calls now start together, moving the critical path closer to the slowest dependency group instead of the sum of each independent call. Wallet support-directory and randomized DB-name results are also cached for the process, so later path consumers avoid repeating platform-channel and secure-storage work. The largest improvement is expected on cold launches with slower keychain or storage access.
- **Large rescans:** the previous loop fetched the latest tip and reopened/queried the wallet DB after every 100-block batch even when there was no pending raw transaction to resubmit. A 340,000-block recovery scan contains about 3,400 such batches, so the new preflight can avoid roughly 3,400 redundant tip RPCs and 3,400 DB opens in the common no-pending-transaction case. The inexpensive predicate now runs through a persistent read-only connection and only enters the heavier tip/resubmission path when a candidate exists.
- **Network-bound sync:** enabled subtree streams and the four transparent-address UTXO streams run concurrently; compact blocks and their preceding tree state are fetched together; and the next compact-block batch is downloaded while the current batch scans. For each independent group, elapsed time should move from roughly the sum of its network waits toward the slowest member. This should matter most over high-latency lightwalletd connections, including Tor and mobile networks.
- **Database work:** transparent outputs from each concurrent group are buffered and committed in one transaction, while the prepared preflight and read-only connection are reused. This removes repeated connection and transaction setup from the batch hot path.
- **Long-tail reliability:** strict range validation prevents empty or short lightwalletd responses from becoming zero-progress loops. Full tip reconciliation prevents stale-tip completion and only rewinds a lower tip when its hash proves divergence. Cancellation and foreground/background handoff now win before post-RPC errors or DB mutations. These changes primarily remove pathological stalls, retries, and incorrect completion states rather than increasing the normal scan rate.
- **Trade-offs:** sync uses bounded additional request concurrency and transient buffering. A height-only tip response can also require an extra `GetBlock` validation RPC. Those costs are deliberate: batch/range limits and stream timeouts bound them, while the validation prevents false completion or destructive rewinds.

## Correctness notes

- Cancellation/mode handoff wins after every awaited tip, fallback-hash, UTXO, and prefetch operation before errors or DB mutations are applied.
- Lower server tips only rewind when the hash at that lower height proves divergence; a matching hash is treated as a lagging server.
- Height-only `GetLatestBlock` responses fall back to `GetBlock`; missing/malformed fallback hashes fail closed.
- Compact-block ranges reject reversed, empty, short, extra, and out-of-order streams.
- Prefetch tasks retain abort-on-drop ownership, including while awaiting their join handle.
- Untrusted subtree indices/heights and refreshed tip heights use checked conversions.

## Validation

- `cargo test --locked --offline wallet::sync_engine:: --lib`: 74 passed.
- `cargo test --locked --offline wallet::sync::transactions::tests --lib`: 52 passed, 0 failed, 2 fixture-gated ignored.
- `cargo check --locked --offline --lib`: passed on the original registry dependency graph; 23 pre-existing warnings.
- `rust/Cargo.toml` and `rust/Cargo.lock` match `chainapsis/main`; this PR has no dependency-source changes.
- Focused Flutter storage/path/bootstrap tests: 49 passed.
- Focused Flutter analyze on changed files/tests: no issues.
- Changed Rust files pass `rustfmt --check`; `git diff --check` passes.

Known base-tree noise:

- Full `cargo fmt --all -- --check` still reports four pre-existing formatting hunks in `ffi.rs` and `network_privacy.rs`; the same hunks reproduce on `chainapsis/main`.
- Full `fvm flutter analyze` reports the same 23 pre-existing unused Ironwood migration warnings; changed-file analysis is clean.
- The full Flutter suite has unrelated existing Ironwood/app-swap failures on the base branch, so the changed startup lanes are the Flutter gate here.

## API surface

No downstream-public Rust, FRB, C FFI, wire, or schema API changes.

Review-relevant internal additions:

- `wallet::sync::has_pending_raw_transaction` — `pub(super)`
- `wallet::sync::transactions::has_pending_raw_transaction` — `pub(super)`
- `wallet::sync_engine::lwd::get_compact_block_hash` — `pub(super)`
- Dart testing seam `resetWalletSupportDirectoryCacheForTesting()` — `@visibleForTesting`

### DIFF
--- a/lib/src/app_bootstrap.dart
+++ b/lib/src/app_bootstrap.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:flutter/foundation.dart' show kDebugMode, visibleForTesting;
@@ -216,7 +217,6 @@ Future<AppBootstrapState> loadAppBootstrap() async {
   try {
     log('bootstrap: loading startup snapshot');
     await ensureIosSecureStoreAccessibilityMigrated();
-    await storage.ensureWalletDbName();
     await _applyE2eBootstrapOverrides(storage);
     var passwordRotationRecoveryFailed = false;
     try {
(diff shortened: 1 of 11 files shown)


ValarDragon:
Closing the umbrella now that the selected work is split into reviewable drafts: #514 (transparent UTXO refresh), #515 → #516 (compact-block validation and scan pipelining), and #517 → #518 (tip identity and resubmission preflight). Startup PR #513 was separately closed after deciding its complexity was not justified by measured gain. The subtree-root concurrency portion is not included in these drafts; this branch remains available if we later choose to revisit it as its own measured change.

PR #513, opened 2026-08-15
Reduce cold-start bootstrap latency

ValarDragon:
Split from #512. This is the independent startup/bootstrap portion; it has no Rust sync-engine dependency.

## What changes

- Starts independent secure-storage, path-provider, and startup-state reads together after the required iOS keychain migration, E2E override, and password-rotation recovery barriers.
- Loads sync status, the active account balance, and recent transaction history concurrently for an unlocked wallet.
- Memoizes the application-support directory and randomized wallet DB name for the process lifetime.
- Invalidates the cached DB name synchronously during wallet reset and serializes the next lookup behind deletion.

## Expected impact

Cold startup currently pays the cumulative latency of several unrelated keychain, filesystem platform-channel, and Rust FFI calls. After this change, those independent operations overlap, so that part of the startup critical path is closer to the slowest operation than to the sum of all operations.

The improvement should be most visible on devices where secure storage, path-provider, or initial SQLite reads have meaningful latency. Repeated DB-path consumers also avoid additional platform-channel and secure-storage round trips after the first successful resolution.

For an unlocked wallet, the first home frame should retain the existing account, balance, and recent-history snapshot without waiting for those three Rust reads serially. Locked-wallet behavior is unchanged: address, balance, and history are not hydrated before unlock.

This PR intentionally makes no benchmark claim and does not change shielded-sync throughput after startup.

## Correctness and tradeoffs

- Required migrations and password-recovery work remain ordering barriers.
- Concurrent results are unwrapped in the legacy serial order, preserving existing error classification when more than one operation fails.
- Failed cache resolutions are evicted and retried; only successful process-lifetime values remain memoized.
- Wallet reset clears the DB-name cache before deletion begins, so a concurrent lookup cannot return the deleted wallet's name.
- `walletExists` is sampled once for the bootstrap snapshot instead of repeatedly.
- The support-directory cache assumes the OS application-support directory remains valid for the process lifetime.

## Automated validation

- `fvm flutter test test/app_bootstrap_test.dart test/core/storage/app_secure_store_test.dart test/core/storage/wallet_paths_test.dart` — 54 passed.
- Focused `fvm flutter analyze` over all six changed files — no issues.
- `git diff --check` — passed.

The tests prove independent reads start before any one completes; endpoint settings begin after their network dependency; reverse completion order preserves field mapping; secure-storage failures retain precedence; Rust snapshot reads overlap and preserve empty-snapshot fallback; DB-name callers share one in-flight lookup, retry failures, and cannot race reset; and support-directory resolution is shared and retryable.

## Validation criteria before merge

- Compare at least 10 cold launches of locked and unlocked wallets against `main`, recording process start to the first usable welcome/unlock/home frame.
- Confirm unlocked startup paints the existing account, balance, and recent history without an empty-state flash.
- Confirm locked startup does not hydrate address, balance, or history.
- Reset a wallet while DB-path consumers are active, then create/import again; verify a new randomized DB filename is used and the deleted path is never reused.
- Verify unavailable secure storage still routes to `/storage-unavailable`.

## Review focus

1. Cache invalidation and lock ordering in `AppSecureStore`.
2. Bootstrap dependency ordering and preserved failure precedence.
3. Safety of concurrent read-only Rust snapshot calls on an unlocked wallet.

## API surface

Adds two Dart-only `@visibleForTesting` seams: `loadAppBootstrapForTesting` and `resetWalletSupportDirectoryCacheForTesting`. `getWalletSupportDirectory()` now has process-lifetime memoization. There are no Rust, FRB, C FFI, wire, schema, or user-facing copy changes.

### DIFF
--- a/lib/src/app_bootstrap.dart
+++ b/lib/src/app_bootstrap.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:convert';
 
 import 'package:flutter/foundation.dart' show kDebugMode, visibleForTesting;
@@ -210,13 +211,28 @@ class AppSyncSnapshot {
   );
 }
 
-Future<AppBootstrapState> loadAppBootstrap() async {
-  final storage = AppSecureStore.instance;
+Future<AppBootstrapState> loadAppBootstrap() {
+  return _loadAppBootstrap(
+    storage: AppSecureStore.instance,
+    getDbPath: _getDbPath,
+  );
+}
 
+@visibleForTesting
+Future<AppBootstrapState> loadAppBootstrapForTesting({
+  required AppSecureStore storage,
+  required Future<String> Function() getDbPath,
+}) {
+  return _loadAppBootstrap(storage: storage, getDbPath: getDbPath);
+}
+
+Future<AppBootstrapState> _loadAppBootstrap({
+  required AppSecureStore storage,
+  required Future<String> Function() getDbPath,
+}) async {
   try {
     log('bootstrap: loading startup snapshot');
     await ensureIosSecureStoreAccessibilityMigrated();
-    await storage.ensureWalletDbName();
     await _applyE2eBootstrapOverrides(storage);
     var passwordRotationRecoveryFailed = false;
     try {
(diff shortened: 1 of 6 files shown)


ValarDragon:
Closing because the implementation complexity and review surface are not justified by measured startup gains. We can revisit a much narrower optimization after real-device profiling identifies the dominant cold-start cost.

PR #524, opened 2026-08-18
Confirm the largest voting share on chain

czarcas7ic:
Before, Vizor scheduled helper shares independently and treated accepted share confirmation as background work, so a completed submission could still have no share on chain.

After, Vizor asks the voting crate to select one largest share across the complete multi-bundle submission, schedules that share for the current time, and keeps the existing submission job running until an accepting helper reports its nullifier in committed chain state. Other shares remain asynchronous, and the wallet-facing API is unchanged.

This temporarily pins the unreleased upstream voting crate revision that owns deterministic selection, scheduling, and restart behavior.


czarcas7ic:
@codex review

PR #526, opened 2026-08-18
Add background vote share recovery

czarcas7ic:
Vizor keeps accepted helper-share checks alive after the vote screen closes and restores pending rounds from durable zcash_voting state on launch, unlock, and resume. Expired or unauthenticated rounds stop before helper requests; lock, account deletion, and wallet reset drain active checks.

Retries use zcash_voting timing and randomized helper ordering, check only helpers still in the current config, stop after the first accepted retry, and do not repeat an ambiguous POST. The branch temporarily pins the coherent 3.0 crate family to the pending-round API revision.

Targeted Flutter and full Rust suites cover restart discovery, ended-round filtering, one-helper retry, config changes, and lifecycle draining. The unchanged helper/chain path was also exercised on a disposable localnet through one committed reveal and a clean restart.

### DIFF
--- a/lib/app.dart
+++ b/lib/app.dart
@@ -96,6 +96,7 @@ import 'src/providers/linux_update_provider.dart';
 import 'src/providers/network_privacy_provider.dart';
 import 'src/providers/rpc_endpoint_failover_provider.dart';
 import 'src/providers/router_refresh_provider.dart';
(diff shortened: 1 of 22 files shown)


czarcas7ic:
@codex reivew

PR #536, opened 2026-08-21
fix(send): preserve consolidation-aware note selection

Thunnini:
## Summary

- forward Zakura's consolidation-aware note query through `ReservedInputSource`
- preserve reserved-note and Orchard migration-lock exclusions while selecting funding and optional consolidation notes
- avoid applying pool-less Orchard migration lock keys to colliding Sapling outputs
- add regression coverage for delegation, re-selection, exclusions, and cross-pool collisions

## Context

PR #530 enabled `NoteSelection::PreferConsolidation`, but ordinary send and fee estimation both query through `ReservedInputSource`. Because that wrapper did not override `select_spendable_notes_for_consolidation`, the trait default fell back to ordinary accumulated-note selection and returned no optional consolidation candidates. Sends continued to work, but the consolidation behavior introduced by #530 was not actually reached.

@p0mvn, could you please check whether that fallback was intentional? This PR assumes the intent of #530 was for ordinary sends to use Zakura's minimum-cardinality funding and opportunistic consolidation while retaining Vizor's reserved-note and migration-lock protections.

## Validation

- `cargo check --locked --all-targets`
- `cargo test --locked --lib` — 670 passed, 4 ignored
- send module tests — 77 passed, 1 existing slow test ignored
- changed-file `rustfmt --check`
- `git diff --check`
- four review/fix passes with no remaining actionable findings

### DIFF
--- a/rust/src/wallet/sync/send.rs
+++ b/rust/src/wallet/sync/send.rs
@@ -62,9 +62,9 @@ use zcash_client_backend::{
             self, create_proposed_transactions, propose_send_max_transfer, propose_shielding,
             ConfirmationsPolicy, TargetHeight,
         },
-        Account as _, AccountMeta, Balance, CoinbaseFilter, InputSource, MaxSpendMode, NoteFilter,
-        NoteRetention, OutputLockStore, ReceivedNotes, TargetValue, TransparentKeyOrigin,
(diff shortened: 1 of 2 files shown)


p0mvn:
LGTM

PR #538, opened 2026-08-21
Submit vote shares concurrently

ValarDragon:
## Summary

- submit all encrypted vote shares concurrently
- retain ordered helper failover and three retries within each share submission
- persist successful delegation records sequentially after network submissions complete
- preserve successful sibling shares when another share is rejected by every helper

This removes the multiplicative delay where each share waited through the same unavailable helper retry window before the next share could start.

## Validation

- focused concurrent share-submission test
- all 129 voting provider tests
- full Flutter test suite: 2,388 passed, 76 skipped
- Flutter analysis on the changed provider and test
- git diff --check

## API surface

No public or pub(crate) API changes.

### DIFF
--- a/lib/src/providers/voting/voting_session_provider.dart
+++ b/lib/src/providers/voting/voting_session_provider.dart
@@ -1590,10 +1590,10 @@ class VotingSessionNotifier extends AsyncNotifier<VotingSessionState> {
         );
       }
 
+      final submissions = <Future<_InitialShareSubmissionResult>>[];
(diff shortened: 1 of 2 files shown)


Thunnini:
Superseded by #539. The new PR includes both commits from this PR: concurrent share submission and safe completion-order persistence with prevalidation. It also adds sibling persistence after an individual DB failure, transient voting-sidecar lock handling, and full desktop Ironwood regtest plus slow-helper E2E coverage. Closing this PR so review and merge can continue on #539.

PR #540, opened 2026-08-21
Honor helper health during share submission

p0mvn:
## Summary
- re-evaluate helper health before each share submission attempt
- deprioritize helpers degraded by concurrent submissions while retaining the all-degraded liveness fallback
- add regression coverage for planner-selected and mid-flight health changes

## Motivation / Why
Initial candidate lists could become stale while concurrent share submissions updated helper health. This allowed a newly degraded helper to remain ahead of healthy alternatives, including when Rust's planner originally placed it first.

## Tests
- `fvm dart format --output=none --set-exit-if-changed lib/src/providers/voting/voting_session_provider.dart test/providers/voting/voting_providers_test.dart`
- `fvm flutter analyze`
- `fvm flutter test`

### DIFF
--- a/lib/src/providers/voting/voting_session_provider.dart
+++ b/lib/src/providers/voting/voting_session_provider.dart
@@ -1690,8 +1690,15 @@ class VotingSessionNotifier extends AsyncNotifier<VotingSessionState> {
     required BigInt submitAt,
   }) async {
     final acceptedServers = <String>[];
-    for (final serverUrl in candidateServers) {
-      if (acceptedServers.length >= targetCount) break;
+    final remainingServers = LinkedHashSet<String>.of(candidateServers);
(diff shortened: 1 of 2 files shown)


p0mvn:
AI will likely complain that there is a risk of submitting to more servers than planned if all are faulty.

But that's the expected behavior: if many servers are down, we should discard the submission policy and aim to get in the share, despite the risk of a single server potentially holding all shares.

i.e., in the case of bad networking conditions, prefer liveness over an additional layer of privacy mitigation.

The shares are still encrypted, and no single server holds all shares, so submission policy to half the servers is merely an additive layer .

Thunnini:
This change was cherry-picked into #544 and is now included in `main` via merge commit e275cb45c6a3dc5f5e996e68c5739e9dc946a523. Closing this PR as handled.

PR #548, opened 2026-08-23
Fix vote failover delay and mobile completion

czarcas7ic:
Transaction confirmation now tries each configured vote server once per polling pass. If every server has a transient failure, the normal polling loop continues; one silent server cannot multiply the 10-second request timeout through nested retries.

At the start of vote casting, Vizor checks all configured helpers concurrently while proof, broadcast, and confirmation work continues. Initial share submission prioritizes helpers that answered healthy, preserves the planned helper count through fallbacks, retries fast transient failures up to the existing three attempts, and moves on after one five-second nonresponse.

Mobile submission also advances from the completed status page when voting was opened through the normal pushed route stack, while leaving any page pushed above it undisturbed.


czarcas7ic:
@codex review

p0mvn (approved):

PR #561, opened 2026-08-24
fix(voting): explain spent nullifier errors

ValarDragon:
Spent-nullifier rejections during delegation (ZKP1) and vote commitment (ZKP2) previously surfaced raw chain diagnostics, including one error per delegation bundle.

Vizor now checks whether the rejected transaction hash already landed. If it did, normal confirmation resumes; otherwise Vizor stops remaining bundle broadcasts and explains that voting started but the submission status could not be recovered. It does not claim another wallet was responsible. Rejections without a transaction hash keep the original diagnostic.

Software retries also reuse persisted ZKP1 setup after an interrupted submission, and PIR cleanup no longer panics while surfacing an earlier preparation failure.


czarcas7ic:
@codex review

czarcas7ic:
@codex review

czarcas7ic:
@codex review

PR #579, opened 2026-08-28
Use Keystone batch signing for wallet sends

czarcas7ic:
Keystone-backed sends and ZEC swap deposits now use zcash-sign-batch even when only one transaction is signed. Vizor retains the complete PCZT, sends the compact signer view to Keystone, and applies the signatures-only result locally, reducing exposed data and the return QR size.\n\nVoting and Ironwood migration now share the same batch request and response helper. TEX and transparent shielding continue using the full-PCZT response because the v1 batch result cannot carry transparent-input signatures.\n\nTransactions requiring more than Keystone's 96-signature batch limit are rejected before the request QR is shown, with guidance to try a smaller amount.


czarcas7ic:
@codex review

PR #586, opened 2026-09-01
Recover confirmed delegation VANs after local state loss

czarcas7ic:
When a delegation submission reports a spent nullifier, Vizor reconciles any known transaction hashes first. Only if that cannot identify the accepted delegation does it ask zcash_voting to discover and persist matching VAN positions. Recovery then stops the active batch and reloads the durable resume plan. The ordinary vote-tree sync validates recovered positions before any ZKP2 proof, and the happy path pays no extra query.

This draft is stacked on the SDK-owned chain submission work and temporarily patches to an integration-only revision combining the independent deterministic-blinding, chain-submission, and recovery branches. The recovery change itself remains based only on its deterministic prerequisite.

### DIFF
--- a/lib/src/providers/voting/voting_service_providers.dart
+++ b/lib/src/providers/voting/voting_service_providers.dart
@@ -533,7 +533,7 @@ abstract interface class VotingRustApi {
     required BigInt operationEpoch,
   });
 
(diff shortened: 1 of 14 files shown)


czarcas7ic:
@codex review

czarcas7ic:
Closing this since we wont get into this edge case with atomic delegation / vote.

PR #590, opened 2026-09-01
Warm Keystone delegation proofs before signing

czarcas7ic:
Keystone accounts now start background ZKP1 work after snapshot bundle preparation. Signing requests reuse the exact persisted transaction while proof generation runs and after restart. Review and submission share hotkey creation through the secure-storage service, including when the review screen closes or its provider container is replaced. Signature storage rejects setup replaced after validation.

This targets roman/next-gen-gov. The SDK pin combines durable transaction reuse and its checks with the feature branch's existing voting observability. Brief setup contention is retried; navigation and action cleanup preserve durable setup. Clients must upgrade before preparing the next round; legacy state repair is outside this change.

Validation: 713 Rust tests and 319 Dart tests passed, plus 1,581 tests on the combined SDK revision. Bindings were regenerated, locked dependency resolution and crate sources were verified, and Rust formatting and targeted Flutter analysis pass. Physical Keystone QR signing remains untested.


rust/Cargo.toml
```diff
@@ -165,17 +164,19 @@ rusqlite = "0.37"
...
 version = "=0.6.0"
-git = "https://github.com/valargroup/zcash_voting.git"
-tag = "v3.1.0-rc.15"
 default-features = false
 features = ["zakura"]
 
+[patch.crates-io]
+zcash_voting = { git = "https://github.com/valargroup/zcash_voting.git", rev = "624c65bb97f4b8def4edb965f9cfb8eadf07559b" }
+vote-commitment-tree = { git = "https://github.com/valargroup/zcash_voting.git", rev = "624c65bb97f4b8def4edb965f9cfb8eadf07559b" }
+vote-commitment-tree-client = { git = "https://github.com/valargroup/zcash_voting.git", rev = "624c65bb97f4b8def4edb965f9cfb8eadf07559b" }
```
czarcas7ic: Patch till zcash_voting 259 merged

czarcas7ic:
@codex review

p0mvn:
AI review comment:

The app-scoped coordinator is necessary with the current API because hotkey creation spans host-owned secure storage (`read -> generate -> write`), and the Rust SDK cannot make that sequence atomic by itself. Moving `_hotkeyEnsures` out of an individual session correctly prevents review warmup and submission from creating different keys, including when the review provider is disposed mid-write.

As a follow-up, consider putting this invariant in the secure-storage abstraction itself—e.g. an atomic/single-flight `VotingHotkeyStore.getOrCreate(accountUuid, roundId, generate, allowCreation)`—with Riverpod only providing that service. That would make every caller inherit the guarantee and avoid tying key uniqueness to one provider/container lifecycle. The SDK should continue validating the durable hotkey binding, while the host keystore owns creation and custody of the secret.

I do not think ordinary SDK-side persistence alone can replace this coordination unless the SDK is given a keystore callback/abstraction or takes ownership of secret storage, which would change the current custody boundary.

PR #591, opened 2026-09-02
Skip wallet history scans for unknown mempool transactions

czarcas7ic:
## Before

Vizor’s mempool observer processes transactions one at a time. For every previously unseen transaction, it first performed a fast indexed lookup in `transactions`. Even when that lookup proved the transaction was not in the database, Vizor then queried `v_transactions` for account mappings.

`v_transactions` is a computed wallet-history report rather than an indexed lookup table. To answer a txid miss, SQLite materialized unions over received notes and spends, joined sent-note, account, and transaction data, grouped the database-wide history, and built temporary B-trees for grouping and `DISTINCT` before it could conclude that the txid was absent.

On a Vizor database containing 4,957 transactions across multiple wallets, this redundant work took a median **1.96 seconds per unrelated mempool transaction**. At the measured rate, a burst of five unrelated transactions would block the serial mempool observer for roughly **9.8 seconds**.

## After

Unknown transactions now return after the indexed lookup. The median lookup fell to **0.775 ms**, or approximately **2,526× faster**. Five unrelated transaction lookups now take roughly **3.9 ms instead of 9.8 seconds**.

Known wallet transactions still use `v_transactions` for account scoping, and unknown transactions still undergo trial decryption, so new inbound payments are not skipped. This removes wasted local database work and mempool-observer delay; it does not claim that the full transaction-processing or historical-sync path is 2,526× faster.

### DIFF
--- a/rust/src/wallet/sync_engine/mempool.rs
+++ b/rust/src/wallet/sync_engine/mempool.rs
@@ -868,6 +868,15 @@ fn lookup_known_pending_tx(db_path: &str, txid_bytes: &[u8]) -> Result<KnownPend
         .map_err(|e| format!("transactions query: {e}"))?
         .is_some();
 
+    // `v_transactions` is derived from `transactions`, so a missing base row cannot have an
+    // account mapping. Avoid materializing the wallet-wide history view for unrelated mempool txs.
+    if !matched {
+        return Ok(KnownPendingTx {
+            matched: false,
+            account_uuids: Vec::new(),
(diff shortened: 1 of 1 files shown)


czarcas7ic:
Where the 1.9s goes:

The code looks like a normal SQL data retrieval, but the v_transactions is a view and not a stored table. Here is the full explanation:

Screenshot 2026-09-01 at 8 35 28 PM

PR #592, opened 2026-09-02
Make synced funds spendable before history enhancement completes

czarcas7ic:
Before, foreground sync waited for every full transaction to be fetched and decrypted before reporting completion, even though compact scanning had already discovered spendable notes. In the exact 4,957-transaction reimport replay, spendability took 698.949 seconds.

After, foreground sync owns a cancellable enhancement worker while compact scanning continues. The final-code replay reached spendable completion in 171.241 seconds, 75.5% sooner, then restored all 4,957 transactions and 9,660 received notes with an empty queue by 877.986 seconds. Full history is slower than the 698.949-second baseline; that tradeoff prioritizes spendability without increasing aggregate GetTransaction concurrency above one.

The worker opens its second connection lazily, remains owned and joined under the global sync guard, and leaves the durable session incomplete until history drains so an interruption resumes safely. Mode 2 remains single-lane.


czarcas7ic:
@codex review

PR #596, opened 2026-09-02
Skip empty resubmission history queries

czarcas7ic:
Before: each sync batch materialized the aggregate transaction-history view to look for pending outbound transactions, even when the base transaction table had no eligible candidate. This became expensive in wallets with thousands of historical transactions.

After: a cheap base-table preflight skips that view in the normal empty case and fails open to the original query if the optimization cannot run. In the exact 4,915-transaction replay, this reduced sync time from 424.8 seconds to 198.6 seconds without increasing fetch concurrency or changing the recovered transactions, raw payloads, notes, or final queue.


czarcas7ic:
@codex review

PR #619, opened 2026-09-04
fix: serialize voting sidecar migrations

Thunnini:
## Summary
- serialize every voting sidecar open with a dedicated per-path migration lock
- keep the migration lock separate from the existing write coordinator to avoid nested-lock deadlocks
- add a concurrent v15-to-v17 migration regression test with integrity verification

## Validation
- `cargo fmt --manifest-path rust/Cargo.toml --check`
- `cargo test --manifest-path rust/Cargo.toml` (732 passed, 4 ignored)
- concurrent v15 migration regression repeated 20 times
- voting module tests: 41 passed


Thunnini:
@codex review

PR #640, opened 2026-09-09
Preserve Gift Card secrets when funding is uncertain

piatoss3612:
Preserve Gift Card recovery secrets when the funding broadcast result is uncertain or a received Card currently has no observed balance. Saved received Cards remain in `Received` after an `Unavailable` result and can be retried; this list behavior still requires a product policy decision.

This PR addresses Codex Security findings about automatic deletion of Gift Card bearer secrets: "Unverified scan absence can delete a funded Gift Card secret" and "Unavailable check can erase the only Gift Card bearer secret." The related product policy decisions and remaining account-deletion issue are documented below.

## Changes

- Sender: remove automatic deletion of ambiguous funding records based only on empty Gift Card transaction history, the main wallet's scan height, or elapsed time. When a funding transaction is found, update the record through the existing submitted/funded states.
- Sender: make `removeUnbroadcastDraft` reject drafts with a recorded submission attempt.
- Receiver: preserve an already saved Card's recovery secret and temporary wallet when initial claim preparation, confirmation refresh, or preparation after mobile receiving-account selection returns `Unavailable`. The Card can be retried from the list.
- Keep temporary-wallet cleanup for new link previews unchanged. No new screens, buttons, backup/removal flows, or UI copy are included.

## Pending policy decisions and remaining issues

- The original non-retention/cleanup behavior was intended to prevent unavailable Cards from lingering in `Received`. The current retention protects against secret loss; it is not an agreed final policy for list visibility.
- Decide whether secret retention should be separate from `Received` visibility, and when retry or removal should be allowed. A Card already spent elsewhere can currently remain in the list.
- An ambiguous sender draft can still block account deletion. This PR does not provide manual recovery-link export or a resolution flow, and the existing review finding remains unresolved.

## Validation

- 50 sender recovery-store/reconciler tests passed.
- 62 desktop Payment Links screen tests passed, including retention of saved unavailable Cards and retry from the list.
- 37 mobile screen and form-factor sanity tests passed, including retention during preparation after receiving-account selection.
- `fvm flutter analyze --no-pub` and `git diff --check` passed.
- Physical-device/native validation was not run. Two separate existing Keystone tests also timed out at the pre-change head; they are not included in the passing counts above.

Reviewable commits: existing automatic-cleanup removal `9ca4fb6f3`, storage deletion guard `e3060e5f3`, and receiver retention `5783ac134`.


lib/src/features/payment_links/screens/payment_links_screen.dart
```diff
@@ -479,20 +479,12 @@ class _PaymentLinksScreenState extends ConsumerState<PaymentLinksScreen> {
...
-    );
-    if (!listed) return;
-    _receivedCards = [
-      for (final record in _receivedCards)
-        if (record.address != link.address) record,
-    ];
-    unawaited(_paymentLinkOperations.forgetReceivedLink(link));
+  /// An empty scan cannot prove that a Card will never receive funds. Keep
+  /// listed Cards and their wallets for retry; new previews remain disposable.
+  Future<void> _releaseUnavailableClaim(PaymentLinkClaimSession session) {
```
piatoss3612: Product policy decision required: this change protects against secret loss; it does not establish the final policy for the `Received` list.

The original behavior intentionally avoided temporary retention or cleaned up unavailable Cards so they would not linger in `Received`. This change preserves an already saved Card's recovery secret and temporary wallet, which also leaves a retryable `Claim` row in `Received` after an `Unavailable` result. That allows recovery when funding arrives late, but a Card already spent elsewhere can also remain in the list.

Before merging, decide whether secret retention should be separate from list visibility and under what conditions retry or removal should be allowed. The current retention behavior should not be treated as the final UX. No new backup screen or removal flow is included.

lib/src/features/payment_links/services/payment_link_recovery_reconciler.dart
```diff
@@ -315,17 +290,12 @@ class PaymentLinkRecoveryReconciler {
...
-          // been mined at and the Gift Card holds nothing: the transaction
-          // never reached the chain.
-          await _store.removeUnbroadcastDraft(address: record.link.address);
-          changed = true;
+          // Absence is not proof that the broadcast failed. The Gift Card is
+          // scanned through separately selected lightwalletd infrastructure,
+          // which can omit compact-block data or disagree with the source
+          // wallet's scan. Retain the bearer secret unless positive funding
+          // evidence promotes it; automatic cleanup could make mined funds
+          // permanently unspendable.
```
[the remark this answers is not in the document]
piatoss3612: This finding remains unresolved. Ambiguous funding secrets are retained and the storage deletion guard is strengthened, but these drafts can still block account deletion. The existing UI provides no manual resolution path.

Recovery-link export and cleanup require a product policy and UX decision. This PR does not add a new backup screen or removal flow, and the PR description explicitly records this limitation. The separate receiver-retention change also has an inline review comment noting that the `Received` visibility policy remains undecided.

lib/src/features/payment_links/screens/payment_links_screen.dart
piatoss3612: @Thunnini

PR #680, opened 2026-09-14
Allow free desktop window resize instead of a locked aspect ratio

### LINKED ISSUE #574: Fix desktop window experience
On a desktop OS where the main window can be resized, only a specific aspect ratio is allowed. And even when it's very big, the content within it is constrained to a very narrow view, despite the window being forced to be very wide.

ayushsingh82:
## Summary

- Fixes #574: the desktop window could only be resized to exactly one of two hardcoded aspect ratios (`windowManager.setAspectRatio()`), so it could never be freely resized to fill a large or unusually-shaped display.
- Drops the aspect-ratio lock in `initializeDesktopWindow` and `AppLayoutNotifier.setMode`, keeping only the minimum-size constraint (`AppLayoutMode.large`/`small`'s `minimumSize`).
- The existing large/small mode inference (`AppLayoutNotifier`'s `WindowListener` callbacks + `_reconcileLayoutWithWindow`) already tolerates arbitrary window ratios for maximize/unmaximize/fullscreen, so responsive layout behavior is unaffected — it now also drives the inference for ordinary user drag-resize, not just OS-level window state changes.

## Before / after

- **Before**: dragging a window edge only ever produced the same fixed 1080:720 (or 65:133) shape; the window could not be made wider or taller without changing both dimensions in lockstep.
- **After**: the window can be freely resized to any shape at or above the minimum size; large/small responsive layout still switches based on the resulting aspect ratio.

## Test plan

- [x] `fvm flutter analyze` — no issues
- [x] `fvm flutter test` — full desktop-lane suite passes
- [ ] Manual verification on Windows/Linux (only verified on macOS locally)


ayushsingh82:

Issue #505, opened 2026-08-12
Support usdttron:address formatting for recipient


ValarDragon:
Some websites use a QR code, that has the address as:
`usdttron:{tron address}`

Can we add a special case formatter that detects that, and would auto-set the correct currency + chain, and read the right address?

PR #681, opened 2026-09-14
Recognize usdttron: prefixed addresses in the swap recipient field

ayushsingh82:
## Summary
- Some exchanges/wallets embed a chain hint ahead of the raw address in withdrawal QR codes, e.g. `usdttron:T9y...` for TRC-20 USDT (#505). Previously the swap/pay destination field treated the whole string as a literal address, producing a malformed address on the wrong chain.
- Adds a small prefix table (`detectSwapPrefixedAddress`) and resolves it in `SwapNotifier.updateDestination`: a recognized prefix switches the composer to the matching supported external asset and strips the prefix from the stored address.
- Covers manual typing, QR scanning, and Pay (shares the same destination path) on both desktop and mobile. Also reorders the "remember this address" save to happen after resolution, so a prefixed address is remembered under its resolved chain instead of verbatim under the previous one.

## Test plan
- [x] Added `test/features/swap/swap_prefixed_address_test.dart`
- [x] `fvm flutter test test/features/swap/swap_prefixed_address_test.dart`


ayushsingh82:

PR #683, opened 2026-09-15
perf(voting): memoize the voting wallet DB path

p0mvn:
Stack 1/4 — base `main`.

Revives part of #560, reimplemented against current `main` (that branch's merge base is 439 commits behind and `voting_session_provider.dart` alone changed ~5,700 lines since, so it had no anchor points left to rebase onto).

## Summary

Voting resolves the wallet DB path per round refresh, tree sync, session start, and readiness poll. Each resolve hits the application support directory (with a `mkdir`) and reads the DB name from secure storage.

Memoize it behind `votingWalletDbPathProvider`.

## Notes

- The resolved name changes only when a wallet reset mints a new one, so the reset funnel in `runWithSyncPausedForWalletReset` clears the cache alongside the sync notifier's own cached path. That funnel is the only call site of `clearCachedWalletDbPath`, so there is no second path to miss.
- A failed resolve is not cached, and concurrent callers share one in-flight lookup.
- The resolver is injectable purely so the cache is unit-testable without platform plugins; production still defaults to `getWalletDbPath`.

This matters more once the later PRs in the stack add a background readiness poll on top of it.

## Tests

- `fvm flutter analyze` — clean
- `fvm flutter test` — 3891 passed, 115 skipped
- New: `test/providers/voting/voting_wallet_db_path_cache_test.dart`

https://claude.ai/code/session_01T3RUNZWzSkJSBcL7GzVPnp


p0mvn:
@codex review

PR #690, opened 2026-09-15
Refresh the mainnet birthday-height interpolation anchor table

### LINKED ISSUE #511: Remove the latency on birthday height selection
Right now picking a birthday height in Vizor requires an RPC call to fetch it.

Instead I suggest we have Vizor build with:
- A block height each day, for the last 3 months
- A block height once per week, for the last 2 years (2024 on)
- A block height once per month, from 2 years ago to genesis

And then take the user's calendar date and map it to the block height preceeding it. (Handling the edge case of say my block height stored is Aug 1, 00 UTC, user selects August 1st, actually sync beginning July 31st block height)

This removes a latency step that adds errors / delays in the critical onboarding path

ayushsingh82:
## Summary
- Fixes #511. `estimateImportBirthdayHeight`'s fast path interpolates a wallet birthday from a small table of deep, immutable mainnet checkpoints, falling back to up to 2 RPC correction probes when the interpolated estimate drifts outside a 6-hour tolerance. The table's last anchor (height 3,000,000) was from July 2025, so any birthday in the most common case — a wallet created recently — fell into a 14+-month-old final segment, the one most likely to need correction probes and their added latency.
- Adds a fresh anchor at height 3,450,000 (2026-08-16 15:27:40 UTC), verified against a public block explorer; the ~75.4s/block implied rate between it and the current tip cross-checks against Zcash's known ~75s target, giving confidence the fetched data is correct.
- Follows the exact existing anchor cadence (round heights spaced roughly 400+ days apart) rather than introducing a new table format.
- A fully RPC-free estimate isn't structurally possible — the current chain tip has to come from the network somehow — but shrinking the stale final segment reduces how often recent-date estimates need correction probes at all.

## Test plan
- [x] `cd rust && cargo test` — existing birthday-estimation tests pass with the new anchor


ayushsingh82:

PR #691, opened 2026-09-15
fix: show the vote amount in the mobile Keystone memo

czarcas7ic:
Reported in Slack: on mobile, a multi-bundle vote's memos showed no amount, while the Keystone showed it correctly.

The memo pager clamped the whole memo to `maxLines: 2`. `zcash_voting`'s `display_memo()` builds `"<sentence>.\nAmount: <x> ZEC."`, and the sentence alone fills both lines on a phone, so the amount never rendered — and because the cut landed on the `\n`, Flutter drew no ellipsis, so there was no hint anything was dropped. The amount is the only part that differs between bundles, which is what the pager is for.

Now only the leading lines are clamped; the final line gets its own budget, so a long round name can't push the amount out or overflow the column. Desktop was never affected — it renders the memo unclamped.

The old test missed this twice: its fixture put a short amount *first* (unlike real memos), and `find.textContaining` matches the widget's string, not the glyphs a clamped paragraph paints. The fixture now uses the real memo shape and the new assertions measure the rendered paragraph.

Verified: new tests fail on the old code and on a simulated re-clamp. Mobile lane has the same 20 pre-existing failures as `main` (all unrelated); desktop lane green.


czarcas7ic:
IMG_1298

Tested here, it now shows

czarcas7ic:
Please note this is not blocking the release and is just a UX improvement.

PR #708, opened 2026-09-16
Preserve Ledger swap and pay results after broadcast

piatoss3612:
## Problem

A Ledger Swap or Pay deposit could be broadcast successfully and then appear as `Ledger signing failed` if proposal cleanup or local completion failed. If cleanup failed before the UI retained the result, `Try again` attempted to broadcast an operation already in `result_pending_ack`, producing a misleading broadcast error.

## Changes

- Preserve accepted or uncertain broadcast results before fallible proposal cleanup. Keep the draft identifiers until settlement succeeds, and retry cleanup before persisting and acknowledging the deposit result.
- Reload durable pending results on retry and resume persistence/acknowledgement without requesting another signature or broadcast. Block duplicate retry clicks while recovery is running.
- Show `Transaction sent` or `Transaction status pending` with `Retry saving` when completion needs recovery. Hide device-app and connection prompts during this stage and display saving progress for persistence-only work.
- Validate both live and recovered results before recording a deposit. Expired transactions never persist their locally computed txid as a deposit or invoke the success callback; retain the draft and durable result until cleanup and acknowledgement succeed, then close through the existing cancellation route. Failed expiry cleanup offers `Retry cleanup`; it never signs or broadcasts again. Background recovery also clears expired outcomes without recording a deposit.
- Give the active signing surface and background recovery mutually exclusive ownership of each operation, so recovery cannot acknowledge an outcome while the surface is retrying draft cleanup. Propagate uncertain-outcome lock-retention errors instead of treating them as successful settlement.
- Keep the signing surface open when user-requested cancellation cannot settle its draft, and protect cancellation cleanup with the existing destructive-operation lifecycle. After successful cleanup, leaving the surface releases ownership and explicitly requests background recovery for any checkpointed operation, including results skipped by an earlier recovery pass.

## Validation

- Desktop deposit regression tests: 13 passed. Shared Ledger modal tests: 11 passed in the earlier validation.
- Ledger operation recovery tests: 12 passed, including expired Swap/Pay outcomes and active-operation ownership exclusion.
- Existing mobile Ledger Swap signing tests: 5 passed with the mobile form-factor define in the earlier validation.
- Scoped Dart analysis and `git diff --check` passed.
- Regression cases cover Swap/Pay persistence retry, post-broadcast cleanup retry without premature acknowledgement, acknowledgement retry, unknown outcomes, durable-result recovery after a broadcast call throws, expired outcomes on initial response, retry, reopening, and background recovery, cleanup/acknowledgement failures after expiry, uncertain-outcome lock-retention failures, and handing a previously skipped result to actual background recovery when leaving the signing surface.

Tests run the production UI/completion service with fault-injected signer, outbox, and persistence services. They do not exercise real Ledger hardware or broadcast live transactions, and do not establish the root cause of the reported RC incident.


piatoss3612:
@codex review

PR #716, opened 2026-09-17
Label self-shields and their receipts as Ironwood once the destination pool is Ironwood

### LINKED ISSUE #414: clarification of ironwood status in app
I see that on the wallet home, it shows "Shielded Balance (Ironwood)" but that is missing from other locations in the UI.  I'd like to see it under Recent Activity for completed transactions as well. "Ironwood Shielded" instead of just "Shielded"  And when you click into the details of an item under Recent Activity, "Ironwood Shielded Balance" instead of just "Shielded Balance"

It would be nice to see something similar on the Send and Receive windows as well.

Given all the miscommunication and multiple sources for verifying the level of ironwood support in wallets, it's disappointing that no wallet has clarified in app what their status is regarding the NU6.3 upgrade.  I would like to see vizor be the first to make it clear in app. 

Additionally, the displaying the status of ironwood for new privacy/shielding transactions would be nice too.  I don't see any  mention of this.  For example, pressing the shield button just results in "shielding" being displayed.  It would be nice if that explained what is happening... "moving to ironwood address"  

Sure, one should assume that what it's doing and you can confirm with the transaction id afterwards, but you know what they say about making an ass of u and me. I think this would be helpful for users that are new to zcash... especially since it is a very user unfriendly ecosystem at this time.

ayushsingh82:
## Summary
- Fixes #414. Home already distinguishes "Shielded balance" from "Shielded balance (Ironwood)" once a wallet's shielded funds have migrated, but Recent Activity, the desktop and mobile transaction-detail receipts, and mobile Home's balance card all showed a bare "Shielded" regardless of which pool a self-shield actually landed in.
- Reuses the existing `display_pool` field already returned per-transaction (and already used for the received/sent subtitle) to label a self-shield as "Shielded (Ironwood)" in Recent Activity and "Shielded balance (Ironwood)" / "Shielded (Ironwood)" in the desktop `ShieldedReceiptView` and its mobile equivalent.
- Brings mobile Home's balance card label to parity with desktop Home, which already had this distinction via `ironwoodHomeBalancePresentationProvider`.
- This is copy-only: `display_pool` and the balance-presentation provider are read-only display concerns already exercised by existing tests. No balance or transaction logic changes.

## Test plan
- [x] `fvm dart format --output=none --set-exit-if-changed <changed files>`
- [x] `fvm flutter analyze`
- [x] `fvm flutter test` (full desktop lane, 4191 passed)
- [x] `fvm flutter test --tags mobile --run-skipped --dart-define=VIZOR_FORM_FACTOR=mobile` (confirmed the only failures are pre-existing on a clean `main` pull, unrelated to this change)


ayushsingh82:

PR #719, opened 2026-09-17
Support a configurable wallet data directory for portable/USB installs

### LINKED ISSUE #145: enhancement : is it possible to define where the config/DB files are used from
i would like to run from a USB NVME drive, in which case i would like to move the settings/DB location to a different location, could it be possible to add a command line argument to instruct the new locations for these files?

ayushsingh82:
## Summary
- Fixes #145. Wallet DB, Tor data, gift-card tracking DBs, and payment-link claim wallets all resolve through `getWalletSupportDirectory()`, which was hardcoded to the OS's per-user application-support directory. A user running Vizor from a USB/NVMe drive had no way to keep wallet data alongside the binary instead.
- Adds `configureWalletDataDirectoryOverride(args)`, called once from `main()` before any wallet path is resolved. It reads a `--wallet-data-dir` CLI argument (already forwarded to Dart's `main()` unmodified by the generated Linux and Windows runners) or the `VIZOR_WALLET_DATA_DIR` environment variable (also covers macOS, where this app's runner does not forward argv to Dart).
- Desktop only; a no-op on mobile, where neither invocation style applies.
- `getWalletSupportDirectory()` uses the override when set and otherwise keeps its original behavior — fully opt-in, no change to any existing install's wallet path.

## Test plan
- [x] `fvm dart format --output=none --set-exit-if-changed <changed files>`
- [x] `fvm flutter analyze`
- [x] `fvm flutter test test/core/storage/wallet_paths_test.dart` (new, 6 cases covering `=`-form, two-token form, unrelated args, empty value, dangling flag, empty arg list)
- [x] `fvm flutter test` (full desktop lane)


ayushsingh82:

Issue #721, opened 2026-09-17
fix: ledger account locked after failed giftcard creation


distractedm1nd:
I got deadlocked today using the new ledger integration.

I started creating a giftcard for 0.1 ZEC, but for some reason it was not able to load the signature request on my ledger. I still dont know the cause of this but it felt like typical ledger bluetooth connection issues on my side, so not necessarily Vizor's fault.

After I gave up however, it showed me I have 0 ZEC balance, and there was also no way to continue the "Funding incomplete" status of the giftcard. In an attempt to clear this, I tried reimporting the wallet. This failed, as I still have pending giftcards.

____

I asked Codex to help me figure out why this was happening. It says:

  1. “Funding incomplete” has no recovery action.
     vizor-wallet/lib/src/features/payment_links/screens/payment_links_screen.dart:2395
     labels saved drafts this way, disables sharing, and provides no resume/cancel
     action. Ledger funding can resume, but that retry is only available inside the
     original signing overlay.

  2. Reserved funds disappear from the displayed shielded balance.
     vizor-wallet/lib/src/providers/sync_provider.dart:329 calculates the displayed
     balance using spendable and pending amounts, omitting locked amounts. Ledger funding
     retains input locks when the outcome is uncertain. If those inputs cover your entire
     balance, the display can become zero even though the funds are reserved. Rust’s
     total balance includes those locks.

  3. The same draft blocks account deletion.
     vizor-wallet/lib/src/features/payment_links/services/
     payment_link_recovery_store.dart:812 counts drafts with a funding transaction ID as
     unshared cards. That ID is saved during proof generation—before signing or
     broadcasting—so this guard can block deletion of an unfunded card. Duplicate-account
     rejection then prevents reimporting as a workaround.

  4. Background recovery has a cleanup gap.
     vizor-wallet/lib/src/features/ledger/services/ledger_operation_recovery.dart:181
     broadcasts directly and only logs failures. On definitive rejection, Rust deletes
     the saved operation, but this path misses the draft cleanup performed by vizor-
     wallet/lib/src/features/payment_links/services/
     payment_link_ledger_funding_service.dart:208. The card can remain stranded awaiting
     expiry reconciliation.

_______

I would try to fix this myself if the stuck client were desktop, but it is mobile, and I assume your prompting will be faster and less blocking than me submitting a PR with no way to test it myself.

distractedm1nd:
[...]

PR #722, opened 2026-09-17
fix: force single change output when using ledger

distractedm1nd:
Related: #721

Doesn't fix the gift card issues but fixes the inability to transact. Tested locally with my ledger and the fix works.

I assume this can be reverted once they allow multiple change outputs upstream!


Thunnini:
@codex review

PR #733, opened 2026-09-18
Fix Ledger voting and exercise real SDK signing requests

piatoss3612:
Stacked on #728.

## Problem

Ledger rejected the SDK's delegation PCZT with `0x6a80` before approval. Its zero-value output pays a separate voting hotkey without an OVK, leaving Ledger with no recoverable output to review. Enabling OVK recovery exposed a second issue. The device app renders a memo as text only when every byte is printable ASCII and otherwise falls back to a memo hash, and we observed a simulator crash on that fallback. The original memo's newline was one way in; any non-ASCII character, tab or control byte is another.

The previous voting fixture was a self-transfer with an empty memo, so it missed both conditions.

## Changes

- Use published `zcash_voting 5.1.0` from crates.io, with the published tree dependencies. Enable `with_ledger_output_review()` only when the account's stored signer type is Ledger; software and Keystone retain the defaults.
- Replace the synthetic voting fixture with two actual SDK-generated signing requests, including persistence and redaction.
- Refuse a memo the device app cannot render as text before the signing request is built, on both the Dart and Rust sides, mirroring the device's own condition. The check runs on the memo the transaction actually pays — empty for a transparent recipient, trimmed otherwise — so a memo that never reaches the device is not refused. The message is worded as a current limitation, since it lifts when the device app is fixed upstream.
- The voting SDK already escapes its delegation memo to printable ASCII, so the guard is a backstop on that path and the active check for sends.
- Give the mobile memo error room to be read. At phone width its slot left 326 logical pixels on one line, so any useful message was clipped — already true before this branch, missed because the tests run at a wider viewport. The slot now holds two lines, taken from the memo text area so the sheet keeps its height, and a test pins the narrow case.
- Add desktop voting coverage through **final tally**, using matching software and Ledger accounts on separate fresh local chains. Fail immediately on a voting error instead of waiting for the completion-screen timeout.

## Validation

| Check | Result |
| --- | --- |
| Desktop software voting → final tally | Passed |
| Desktop Ledger simulator voting → final tally | Passed: simulator-generated signature, proof generation, chain submission and tally |
| Exact tally comparison | Both runs included all 64 shares. Each of four proposals had **1 ballot for decision 0 and 0 for the other decision**. |
| Standalone Ledger signing | Desktop and Android passed two sequential SDK-generated requests in earlier runs. |
| Rust regression tests | Voting: 35 passed; accounts: 45 passed. |
| Published SDK update | Locked metadata and dependency trees passed. All 341 Rust source files across the three published voting crates match the previous fork. |
| Memo guard | Desktop Dart 654 passed; mobile lane 206 passed; Rust 880 passed. Rust cases cover non-ASCII, tab, CR/LF, invalid UTF-8, a non-text lead byte, the ZIP-302 "no memo" form and a 512-byte ASCII memo. |
| Memo sheet geometry | Measured at 375x667, 390x844 and 430x932, with and without the keyboard. Sheet height is unchanged from before this branch. |
| Changed test/tooling checks | Dart analysis, Rust helper build/formatting, shell syntax and Python compilation passed. |

The earlier non-Ledger Android voting run failed **before voting**, during lightwalletd sync (`ENHANCE_YOUR_CALM / too_many_data_frames`). It does not establish Android voting compatibility; final-tally coverage added here is desktop-only.

## Reproduce

```bash
scripts/e2e/ledger-speculos-docker.sh voting-tally
```

This runs both account types. Set `E2E_LEDGER_VOTING=true` to rerun only Ledger. Setup and artifact details: [Speculos documentation](https://github.com/chainapsis/vizor-wallet/blob/38c38f56c90bf89ae4856a841036dc42a6937a63/docs/ledger/speculos.md#desktop-voting-through-final-tally).

## Scope and limitations

- Final-tally tests use a **custom testnet build** of the pinned Ledger Zcash app: `testnet` feature plus coin-type-1 install paths. The distributed app is mainnet-only; Vizor's production network restriction remains unchanged.
- A test-only signer provider calls the existing APDU transport and signature verification through a local helper. This bypasses the product API's mainnet account gate. SDK delegation, proofs, submission and tally are real; existing mainnet signing tests cover the product wrapper separately.
- Funding/migration uses a software fixture before importing the device UFVK. This does not test Ledger shielding, physical devices or Stax.
- The SDK option lets account-OVK holders recover the voting-hotkey output and delegation memo. Existing persisted PCZTs are reused unchanged.
- The memo guard is a client-side measure while the device app is fixed upstream, not a fix. It does not model the device's per-transaction memo retention budget, which takes three maximum-length memos to reach; our proposals carry at most one memo-bearing output.
- Measuring the sheet surfaced two layout problems that predate this branch and are left alone: the send review screen overflows a button label at phone width (further for a Keystone account, which never reaches this guard), and on a 375x667 device with the keyboard up the memo sheet is 97 logical pixels taller than the space it gets.


Thunnini:
@codex review

